PR #34, opened 2026-03-30
feat: add scion images build command

fijimunkii:
## Summary

- Adds `scion images build` command that wraps `image-build/scripts/build-images.sh`, so users can build container images directly from the CLI without needing to find and invoke the script manually
- Supports all the same flags as the script: `--registry` (required), `--target`, `--push`, `--platform`, `--tag`, plus `--source` to specify the repo clone path when not running from within it
- Exempts the `images` command group from grove and registry requirements in `PersistentPreRunE` — it's used to set up the registry before those things exist

## Test plan

- [x] Unit tests added covering: command/flag registration, missing script error, correct arg passthrough with and without optional flags
- [x] Manually verified end-to-end: `scion images build` correctly invokes docker buildx with expected arguments
- [x] Pre-existing test failures in `pkg/agent` and `pkg/runtimebroker` confirmed unrelated to this change

> **Note:** Depends on fix for `build-images.sh` empty platform array bug to work without `--platform`.

### DIFF
--- a/cmd/images.go
+++ b/cmd/images.go
@@ -1,0 +1,129 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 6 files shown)


ptone:
this has been reworked into each harness providing a portable docker file and a system to build the external authors harnesses

PR #55, opened 2026-04-08
Add provider selection for hub auth login

mfreeman451:
## Summary
- add `--provider` to `scion hub auth login`
- add a hub auth providers endpoint so the CLI can discover configured providers for browser and device flows
- auto-select the provider only when exactly one is configured, and require an explicit choice when multiple providers are available

## Problem
CLI auth currently does not have a clean way to target a specific OAuth provider. That causes two user-facing issues:
- a hub configured for GitHub-only auth can fail headless login when the CLI assumes a different provider
- a hub configured with multiple providers has no explicit selection path for CLI login

## Approach
This keeps provider selection explicit and predictable instead of relying on client-side fallback behavior:
- `scion hub auth login --provider <provider>` lets users choose directly
- when `--provider` is omitted, the CLI asks the hub which providers are configured for the relevant client flow
- if exactly one provider is configured, the CLI uses it automatically
- if multiple providers are configured, the CLI returns an error asking the user to pass `--provider`

## Testing
- `go test ./cmd ./pkg/hub/auth ./pkg/hubclient`
- `go test ./pkg/hub -run 'TestCLIAuthProviders|TestCLIAuthProviders_InvalidClientType|TestCLIDeviceAuthorize|TestCLIDeviceToken|TestProviderInferenceFromGitHubRedirectURI'`

### DIFF
--- a/cmd/hub_auth.go
+++ b/cmd/hub_auth.go
@@ -34,23 +34,16 @@ var (
 	hubAuthNoBrowser bool
 )
 
-const (
-	hubAuthProviderGoogle = "google"
-	hubAuthProviderGitHub = "github"
-	hubAuthClientTypeCLI  = "cli"
-	hubAuthClientTypeDev  = "device"
-)
-
 type invalidHubAuthProviderError struct {
 	provider string
 }
 
 func (e invalidHubAuthProviderError) Error() string {
-	return fmt.Sprintf("invalid provider %q: must be google or github", e.provider)
+	return fmt.Sprintf("invalid provider %q: must be one of %s", e.provider, strings.Join(hubclient.OAuthProviderOrder(), ", "))
 }
 
 type noHubAuthProvidersConfiguredError struct {
-	clientType string
+	clientType hubclient.OAuthClientType
 }
 
 func (e noHubAuthProvidersConfiguredError) Error() string {
@@ -58,7 +51,7 @@ func (e noHubAuthProvidersConfiguredError) Error() string {
 }
 
 type multipleHubAuthProvidersConfiguredError struct {
-	clientType string
+	clientType hubclient.OAuthClientType
 	providers  []string
 }
 
@@ -143,7 +136,7 @@ func runHubAuthLogin(cmd *cobra.Command, args []string) error {
 	}
 
 	if hubAuthNoBrowser || util.IsHeadlessEnvironment() {
-		provider, err := resolveHubAuthProvider(cmd.Context(), client.Auth(), hubAuthClientTypeDev, provider)
+		provider, err := resolveHubAuthProvider(cmd.Context(), client.Auth(), hubclient.OAuthClientTypeDevice, provider)
 		if err != nil {
 			return err
 		}
@@ -166,7 +159,7 @@ func runHubAuthLogin(cmd *cobra.Command, args []string) error {
 
 // runBrowserAuthFlow performs the browser-based OAuth flow.
 func runBrowserAuthFlow(cmd *cobra.Command, client hubclient.Client, requestedProvider string) (*hubclient.CLITokenResponse, error) {
-	provider, err := resolveHubAuthProvider(cmd.Context(), client.Auth(), hubAuthClientTypeCLI, requestedProvider)
+	provider, err := resolveHubAuthProvider(cmd.Context(), client.Auth(), hubclient.OAuthClientTypeCLI, requestedProvider)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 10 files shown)


mfreeman451 (commented):
lgtm

ptone:
Essentially looks fine to me - I'd consider the AI provided feedback below optional - and maybe something that can be addressed in your future OIDC proposal

## Summary

This branch introduces explicit OAuth provider selection for hub auth, adding a `--provider` flag, a provider discovery endpoint (`/api/v1/auth/providers`), and auto-selection logic. The change is well-structured with solid test coverage.

---

## Issues

### 1. Hardcoded fallback in `NewDeviceFlowAuth` — silent default could mask misconfiguration

**File:** `pkg/hub/auth/device_flow.go:37-39`

```go
if provider == "" {
    provider = "google"
}
```

The caller (`runHubAuthLogin`) already resolves the provider via `resolveHubAuthProvider`, which errors on empty. This fallback silently masks a bug if someone calls `NewDeviceFlowAuth` with an empty provider from a different call site. Consider either removing the fallback (let it fail loudly) or documenting why it's intentional. At minimum, this creates a coupling between the default here and whatever the hub actually has configured.

### 2. `--provider` flag registered as a local string, not bound to a package var

**File:** `cmd/hub_auth.go:117`

```go
hubAuthLoginCmd.Flags().String("provider", "", "OAuth provider to use (google or github)")
```

All other flags in this file use `StringVar` bound to a package-level var (`hubAuthHubURL`, `hubAuthNoBrowser`). This one uses `String` and is read later via `cmd.Flags().GetString("provider")`. This inconsistency is minor but could cause confusion. Consider aligning with the existing pattern.

### 3. Provider constants are duplicated across packages

**Files:** `cmd/hub_auth.go:37-38` and `pkg/hub/oauth.go:95-96`

`cmd/hub_auth.go` defines:
```go
hubAuthProviderGoogle = "google"
hubAuthProviderGitHub = "github"
```

`pkg/hub/oauth.go` defines:
```go
googleOAuthProvider = "google"
githubOAuthProvider = "github"
```

These are the same values in two places. If a third provider is added, both must be updated. Consider exporting the constants from one canonical location (the `oauth.go` package seems natural) and importing them in `cmd/`.

### 4. `AuthProvidersResponse` type duplicated between hub and hubclient

**Files:** `pkg/hub/handlers_auth.go:119-123` and `pkg/hubclient/auth.go:96-101`

`CLIAuthProvidersResponse` (hub) and `AuthProvidersResponse` (hubclient) are structurally identical. This is a common Go pattern for API boundaries, so it's not a bug, but worth noting for maintainability — if the response shape changes, both must be updated.

### 5. `clientType` parameter naming inconsistency

**Files:** `cmd/hub_auth.go:36-37` and `pkg/hub/oauth.go:99-103`

On the client side:
```go
hubAuthClientTypeCLI  = "cli"
hubAuthClientTypeDev  = "device"
```

On the server side:
```go
OAuthClientTypeCLI    OAuthClientType = "cli"
OAuthClientTypeDevice OAuthClientType = "device"
```

The client uses `hubAuthClientTypeDev` — note the shortened name `Dev` vs `Device`. This naming inconsistency could cause confusion when reading the code.

---

## Minor / Style

- The mock in `cmd/hub_auth_test.go` (`mockHubAuthNotImplementedError`) duplicates the identical pattern in `pkg/hub/auth/device_flow_test.go` (`mockDeviceFlowNotImplementedError`). If the `AuthService` interface grows, maintaining these mocks separately becomes a burden. A shared test mock or a mock generator would help.

- `resolveHubAuthProvider` validates against only `"google"` and `"github"` using string constants. The error message also hardcodes `"must be google or github"`. If the set of valid providers grows, this validation should derive from the canonical list (e.g., `oauthProviderOrder()`).

---

## What Looks Good

- **The `resolveHubAuthProvider` logic is clean**: explicit provider skips discovery, single provider auto-selects, multiple providers errors with a helpful message. Good UX.
- **Test coverage is thorough**: all branches of `resolveHubAuthProvider` are tested, device flow tests verify provider propagation, and the server handler tests cover both happy path and invalid input.
- **The `/api/v1/auth/providers` endpoint** is correctly marked as unauthenticated (it needs to be callable before login).
- **`ConfiguredProvidersForClient` uses stable ordering** via `oauthProviderOrder()`, and capacity is correctly derived from the order slice.
- **The `omitempty` on the provider JSON fields** preserves backward compatibility with hubs that don't yet support provider selection.

---

## Recommendation

The most actionable items are **#1** (silent default) and **#3** (duplicated constants). The rest are style/maintainability observations.

mfreeman451 (commented):
lgtm

PR #62, opened 2026-04-08
Support in-cluster Kubernetes auth

mfreeman451:
## Summary
- fall back to `rest.InClusterConfig()` when no explicit kubeconfig path or context is provided
- preserve existing behavior when a kubeconfig path or context is explicitly set
- add focused tests for the fallback and non-fallback paths

## Why
Scion's Kubernetes runtime currently assumes kubeconfig loading even when the broker is running inside Kubernetes. In-cluster deployments should use the mounted service account credentials by default instead of requiring a separate kubeconfig artifact.

## Validation
- `go test ./pkg/k8s`
- `golangci-lint run --config .golangci.yml --new-from-rev=main ./pkg/k8s`
- `make build`
- in-cluster validation against a real Scion deployment confirmed the new code path uses in-cluster config successfully once Kubernetes API egress is allowed by policy


mfreeman451 (commented):
lgtm

PR #71, opened 2026-04-08
Fix OAuth provider defaulting to Google when only GitHub is configured

### LINKED ISSUE #69: CLI and server default OAuth provider to Google with no way to override
## Problem

When configuring a Scion Hub with only GitHub as an OAuth provider (no Google), the CLI `hub auth login` and `hub auth login --no-browser` commands fail because both the server and client hardcode Google as the default provider.

### Server side

In `pkg/hub/handlers_auth.go`, five locations default `provider = "google"` when the client doesn't specify one (lines 324, 753, 812, 956, 1010):

```go
provider := req.Provider
if provider == "" {
    provider = "google"
}
```

### Client side

In `pkg/hub/auth/device_flow.go`, the device flow hardcodes Google:

```go
codeResp, err := d.client.RequestDeviceCode(ctx, "google")
// ...
pollResp, err := d.client.PollDeviceToken(ctx, codeResp.DeviceCode, "google")
```

### Result

With only GitHub OAuth configured:

```
$ scion hub auth login
Error: failed to get auth URL: validation_error: OAuth provider not configured for CLI: google (status: 400)

$ scion hub auth login --no-browser
Error: device flow authentication failed: failed to request device code: validation_error: OAuth provider not configured for device flow: google (status: 400)
```

There is no `--provider` flag on `hub auth login` and no client-side config to change the default.

## Suggested fix

**Server**: When the requested provider is empty and the default (Google) isn't configured, fall back to whichever provider *is* configured. For example:

```go
if provider == "" {
    provider = "google"
    if !s.oauthService.IsProviderConfiguredForClient(clientType, provider) {
        provider = "github"
    }
}
```

**Client**: Either add a `--provider` flag to `hub auth login`, or use the same fallback logic — query the server for available providers before requesting auth.

meatballs:
## Summary
- Server: add `DefaultProviderForClient` to `OAuthService` that checks which providers are actually configured, preferring Google for backwards compatibility but falling back to GitHub
- Client: pass empty provider in device flow requests, letting the server choose the default via `omitempty` JSON serialization
- Applies to all five handler locations in `handlers_auth.go` plus the two client-side calls in `device_flow.go`

Fixes #69

### DIFF
--- a/pkg/hub/handlers_auth.go
+++ b/pkg/hub/handlers_auth.go
@@ -324,14 +324,17 @@ func (s *Server) handleAuthToken(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	// Default provider to google for now if not specified in request
+	// Default provider if not specified in request
 	provider := req.Provider
 	if provider == "" {
-		provider = "google"
 		if strings.Contains(req.RedirectURI, "github") {
 			provider = "github"
+		} else if s.oauthService != nil {
+			provider = s.oauthService.DefaultProviderForClient(OAuthClientTypeWeb)
+		} else {
+			provider = "google"
 		}
-		slog.Debug("OAuth provider inferred from redirect URI", "provider", provider)
+		slog.Debug("OAuth provider inferred", "provider", provider)
 	}
 
 	// Validate provider is a known value
@@ -753,10 +756,14 @@ func (s *Server) handleCLIAuthAuthorize(w http.ResponseWriter, r *http.Request) 
 		return
 	}
 
-	// Default to Google if no provider specified
+	// Default provider if not specified
 	provider := req.Provider
 	if provider == "" {
-		provider = "google"
+		if s.oauthService != nil {
+			provider = s.oauthService.DefaultProviderForClient(OAuthClientTypeCLI)
+		} else {
+			provider = "google"
+		}
 	}
 
 	// Check if OAuth service is configured
@@ -854,10 +861,14 @@ func (s *Server) handleCLIAuthToken(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	// Default to Google if no provider specified
+	// Default provider if not specified
 	provider := req.Provider
 	if provider == "" {
-		provider = "google"
+		if s.oauthService != nil {
+			provider = s.oauthService.DefaultProviderForClient(OAuthClientTypeCLI)
+		} else {
+			provider = "google"
+		}
 	}
 
 	// Check if OAuth service is configured
@@ -1001,7 +1012,11 @@ func (s *Server) handleCLIDeviceAuthorize(w http.ResponseWriter, r *http.Request
 
 	provider := req.Provider
 	if provider == "" {
-		provider = "google"
+		if s.oauthService != nil {
+			provider = s.oauthService.DefaultProviderForClient(OAuthClientTypeDevice)
+		} else {
+			provider = "google"
+		}
 	}
 
 	if s.oauthService == nil {
@@ -1055,7 +1070,11 @@ func (s *Server) handleCLIDeviceToken(w http.ResponseWriter, r *http.Request) {
 
 	provider := req.Provider
 	if provider == "" {
-		provider = "google"
+		if s.oauthService != nil {
+			provider = s.oauthService.DefaultProviderForClient(OAuthClientTypeDevice)
+		} else {
+			provider = "google"
+		}
 	}
 
 	if s.oauthService == nil {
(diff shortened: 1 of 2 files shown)


ptone:
# PR #71 Review: Fix OAuth provider defaulting to Google when only GitHub is configured

__AI code review__

## Executive Summary

This PR replaces hardcoded `"google"` provider defaults across five OAuth handler endpoints with a dynamic lookup via a new `DefaultProviderForClient` method, fixing the case where only GitHub is configured. **One handler (`handleAuthToken`) has a client-type mismatch bug that could select the wrong default provider for CLI callers, and the PR description claims client-side changes in `device_flow.go` that are not present in the diff.**

## Critical Issues

### 1. Wrong `OAuthClientType` used for default provider in `handleAuthToken`

**File:** `pkg/hub/handlers_auth.go`, lines ~327-336 (new code)

The new code in `handleAuthToken` hardcodes `OAuthClientTypeWeb` when looking up the default provider:

```go
} else if s.oauthService != nil {
    provider = s.oauthService.DefaultProviderForClient(OAuthClientTypeWeb)  // Always Web!
}
```

However, the actual `oauthClientType` is determined **after** the provider default, on lines 344-350:

```go
clientType := ClientTypeCLI          // default is CLI
oauthClientType := OAuthClientTypeCLI
if strings.ToLower(req.ClientType) == "web" {
    clientType = ClientTypeWeb
    oauthClientType = OAuthClientTypeWeb
}
```

If a CLI client calls this endpoint without specifying a provider, it will get the default provider for the **Web** client type — which may differ from the CLI configuration. For example, if Web is configured with Google only and CLI with GitHub only, a CLI caller would incorrectly default to Google and then fail the `IsProviderConfiguredForClient` check downstream.

**Suggested Fix:** Move the client-type resolution above the provider-defaulting block, or parse `req.ClientType` inline:

```go
provider := req.Provider
if provider == "" {
    if strings.Contains(req.RedirectURI, "github") {
        provider = "github"
    } else if s.oauthService != nil {
        // Determine client type first to select correct default
        ct := OAuthClientTypeCLI
        if strings.ToLower(req.ClientType) == "web" {
            ct = OAuthClientTypeWeb
        }
        provider = s.oauthService.DefaultProviderForClient(ct)
    } else {
        provider = "google"
    }
    slog.Debug("OAuth provider inferred", "provider", provider)
}
```

### 2. PR description/diff mismatch — missing client-side changes

The PR body states:

> Client: pass empty provider in device flow requests, letting the server choose the default via `omitempty` JSON serialization
> Applies to ... the two client-side calls in `device_flow.go`

However, `device_flow.go` is not in the changed files list. Either the client-side changes were accidentally dropped from the branch, or the description is stale. If the client still sends `"google"` as the provider, the server-side fallback logic added here will never be reached for device flow callers.

**Action needed:** Confirm whether the client-side changes are intentionally omitted or missing.

## Observations

### 3. Duplicated defaulting pattern (minor)

The same nil-check + `DefaultProviderForClient` + `"google"` fallback block is copy-pasted across four handlers (`handleCLIAuthAuthorize`, `handleCLIAuthToken`, `handleCLIDeviceAuthorize`, `handleCLIDeviceToken`). Consider extracting a helper:

```go
func (s *Server) defaultProvider(clientType OAuthClientType) string {
    if s.oauthService != nil {
        return s.oauthService.DefaultProviderForClient(clientType)
    }
    return "google"
}
```

This would reduce the 5-line block to a single call and make future changes less error-prone.

### 4. `DefaultProviderForClient` fallback is redundant but safe

`DefaultProviderForClient` returns `"google"` when no providers are configured, and every call site also has its own `"google"` fallback for when `oauthService` is nil. The double fallback is harmless but worth noting — the `"google"` in `DefaultProviderForClient` is the operative one when the service exists but has no providers configured, while the call-site fallback covers the nil-service case.

### 5. `handleAuthToken` retains redirect-URI sniffing

The first hunk in `handleAuthToken` still checks `strings.Contains(req.RedirectURI, "github")` before consulting `DefaultProviderForClient`. This heuristic is fragile (any redirect URI containing "github" in the path would match), but it is **pre-existing** behavior preserved by this PR. No action required in this PR, but worth a follow-up to deprecate this heuristic now that proper provider detection exists.

## Positive Feedback

- **`DefaultProviderForClient` is clean and well-placed.** Delegating to `ConfiguredProvidersForClient` (which respects `oauthProviderOrder()`) ensures Google is preferred for backward compatibility while correctly falling back to GitHub when it's the only configured provider.
- **Nil-guard on `oauthService`** is consistently applied at all five call sites, preventing panics when OAuth is not configured.
- **The fix directly addresses the reported issue (#69)** with minimal surface area — the new method is only 5 lines and the handler changes are mechanical.

## Final Verdict

**Needs Rework.** The client-type mismatch in `handleAuthToken` (Issue #1) is a correctness bug that could cause auth failures for CLI clients using that endpoint when Web and CLI have different provider configurations. The missing client-side changes (Issue #2) should also be clarified before merge.

ptone:
replaced by #135

PR #72, opened 2026-04-08
Allow broker to start without HMAC keys for registration bootstrap

### LINKED ISSUE #70: Broker registration chicken-and-egg problem
## Problem

When registering a remote broker with a hub, there's a circular dependency:

1. `scion broker start` requires HMAC auth keys to connect to the hub (fails with "runtime broker hub mode requires HMAC auth keys, but none are available")
2. `scion broker register` requires the broker to be running on port 9800 (fails with "broker server not running on port 9800")

HMAC keys are only obtained during registration, but registration checks that the broker is healthy first.

### Additional constraint

Even with `--no-hub`, binding to `0.0.0.0` (production mode) fails with "runtime broker API bound to 0.0.0.0 requires strict broker auth with valid HMAC keys".

### Error sequence

```
$ scion broker start
Error: runtime broker server error: runtime broker hub mode requires HMAC auth keys, but none are available

$ scion broker register
Error: broker server not running on port 9800.
Start it with: scion broker start
```

## Suggested fix

Either:

1. **Allow `broker register` without a running broker** — registration only needs to talk to the hub API and save credentials locally. The local health check is unnecessary for this step.

2. **Allow `broker start` without HMAC keys when unregistered** — start in a degraded state that accepts registration but rejects agent dispatches until HMAC keys are obtained.

3. **Combine both into a single command** — `scion broker start --register` that handles the full bootstrap sequence: start locally, register with hub, obtain keys, restart with full auth.

meatballs:
## Summary
- Allow hub-connected brokers to start without HMAC auth keys, logging a warning instead of failing
- After `scion broker register` obtains the keys, a restart brings the broker to full auth
- This breaks the circular dependency where `broker start` needs keys and `broker register` needs a running broker

Fixes #70

### DIFF
--- a/pkg/runtimebroker/hub_connection_test.go
+++ b/pkg/runtimebroker/hub_connection_test.go
@@ -862,7 +862,7 @@ func TestRuntimeBroker_DefaultAuth_DeniesUnauthenticatedRequests(t *testing.T) {
 	}
 }
 
-func TestValidateBrokerAuthStartup_HubModeWithoutKeysFails(t *testing.T) {
+func TestValidateBrokerAuthStartup_HubModeWithoutKeysAllowsBootstrap(t *testing.T) {
 	cfg := DefaultServerConfig()
 	cfg.Host = "127.0.0.1"
 	cfg.HubEnabled = true
@@ -870,8 +870,8 @@ func TestValidateBrokerAuthStartup_HubModeWithoutKeysFails(t *testing.T) {
 	cfg.InMemoryCredentials = nil
 
 	srv := New(cfg, &mockManager{}, &runtime.MockRuntime{})
-	if err := srv.validateBrokerAuthStartup(); err == nil {
-		t.Fatal("expected startup validation to fail when hub mode has no HMAC keys")
+	if err := srv.validateBrokerAuthStartup(); err != nil {
+		t.Fatalf("expected startup validation to succeed for registration bootstrap, got: %v", err)
 	}
 }
 
--- a/pkg/runtimebroker/server.go
+++ b/pkg/runtimebroker/server.go
@@ -629,9 +629,16 @@ func (s *Server) validateBrokerAuthStartup() error {
 	hasKeys := s.authKeyCount() > 0
 	loopbackOnly := isLoopbackHost(s.config.Host)
 
-	// Hub-connected brokers must have usable HMAC keys for inbound request auth.
+	// Hub-connected brokers without HMAC keys are allowed to start in a
+	// degraded state so that 'scion broker register' can complete the
+	// bootstrap. Without this, the broker cannot start (needs keys) and
+	// cannot register (needs running broker).
 	if s.config.HubEnabled && !hasKeys {
-		return fmt.Errorf("runtime broker hub mode requires HMAC auth keys, but none are available")
+		slog.Warn("Runtime Broker starting without HMAC auth keys — run 'scion broker register' then restart",
+			"host", s.config.Host,
+			"hubEnabled", s.config.HubEnabled,
+		)
+		return nil
 	}
 
 	// Non-loopback listeners must not run without strict broker auth and keys.


ptone:
# PR #72 Review: Allow broker to start without HMAC keys for registration bootstrap

## Executive Summary

This PR relaxes broker startup validation to break a chicken-and-egg bootstrap cycle, but the implementation introduces a **critical security bypass**: the early `return nil` skips the non-loopback safety guard, allowing a hub-enabled broker bound to a public interface (e.g., `0.0.0.0`) to start without any authentication at all. The intent is sound, but the control flow needs rework before merging.

---

## Critical Issues

### 1. Security Bypass: Non-loopback broker can start without auth (Severity: Critical)

**File:** `pkg/runtimebroker/server.go`, lines 633-640 (post-patch)

The new code path returns `nil` immediately when `HubEnabled && !hasKeys`, which **bypasses** the subsequent non-loopback safety check on line 638:

```go
// NEW (lines 633-640): returns nil, skipping everything below
if s.config.HubEnabled && !hasKeys {
    slog.Warn("Runtime Broker starting without HMAC auth keys — ...")
    return nil  // <-- EARLY RETURN
}

// THIS CHECK IS NOW UNREACHABLE when HubEnabled && !hasKeys:
if !loopbackOnly && (!strictAuthConfigured || !hasKeys) {
    return fmt.Errorf("runtime broker API bound to %q requires strict broker auth ...", s.config.Host)
}
```

**Impact:** A broker configured with `HubEnabled=true`, `Host="0.0.0.0"`, and no HMAC keys will now start successfully on all network interfaces with zero authentication. The `brokerAuthMiddleware` is explicitly set to `nil` when there are no keys (line 585-586), so there is no runtime auth fallback either.

**Exploitation scenario:** An attacker on the local network can reach the broker API on any interface — agent creation, message relay, and runtime operations would all be unprotected.

**Suggested Fix:** Remove the early return. Log the warning and fall through so the remaining guards still apply. Only loopback-bound brokers should be allowed to enter the degraded bootstrap state:

```go
if s.config.HubEnabled && !hasKeys {
    if !loopbackOnly {
        return fmt.Errorf("runtime broker hub mode on non-loopback host %q requires HMAC auth keys; "+
            "bind to 127.0.0.1 for registration bootstrap", s.config.Host)
    }
    slog.Warn("Runtime Broker starting without HMAC auth keys — run 'scion broker register' then restart",
        "host", s.config.Host,
        "hubEnabled", s.config.HubEnabled,
    )
    return nil
}
```

### 2. Missing test coverage for the bypass scenario

**File:** `pkg/runtimebroker/hub_connection_test.go`

The existing test `TestValidateBrokerAuthStartup_NonLoopbackPermissiveModeFails` (line 878) uses a config **with** credentials, so it never exercises the newly opened path. There is no test that asserts a non-loopback, hub-enabled, keyless broker is still rejected.

**Suggested Fix:** Add a test case:

```go
func TestValidateBrokerAuthStartup_HubModeNonLoopbackWithoutKeysFails(t *testing.T) {
    cfg := DefaultServerConfig()
    cfg.Host = "0.0.0.0"
    cfg.HubEnabled = true
    cfg.HubEndpoint = "http://hub.example.com:9810"
    cfg.InMemoryCredentials = nil

    srv := New(cfg, &mockManager{}, &runtime.MockRuntime{})
    if err := srv.validateBrokerAuthStartup(); err == nil {
        t.Fatal("expected startup validation to fail for non-loopback hub broker without HMAC keys")
    }
}
```

---

## Observations

1. **Log message uses an em-dash (`—`):** Minor, but structured log consumers may prefer ASCII. Consider `--` or `:` instead.

2. **Test rename is accurate:** `TestValidateBrokerAuthStartup_HubModeWithoutKeysAllowsBootstrap` correctly reflects the new expected behavior (assuming the loopback-only fix is applied).

3. **The `return nil` pattern is fragile:** Future modifications to `validateBrokerAuthStartup` that add checks below the hub block would also be silently bypassed. A fall-through design (or restructuring into explicit `switch`-like guard clauses with a final return) would be more resilient to future changes.

---

## Positive Feedback

- **Problem statement is well-articulated:** The PR description and code comments clearly explain the bootstrap circular dependency being solved.
- **Minimal diff:** The change is scoped tightly to the validation function and its test, making it easy to review.
- **Structured logging:** The warning includes relevant context fields (`host`, `hubEnabled`), which aids operational debugging.

---

## Final Verdict

**Needs Rework.** The early `return nil` creates a security bypass that allows unauthenticated brokers on public interfaces. The fix is straightforward (gate the permissive path on `loopbackOnly`) and should include a new test for the non-loopback + hub + no-keys scenario. Once addressed, the change is safe to merge.

ptone:
I decided to patch this in a way that keeps the security clean, thanks for the issue, closed in 2cdef3e and c03509114360de92d66fc7e814382157d29247d8

PR #75, opened 2026-04-08
Harden Kubernetes agent pod security

mfreeman451:
## Summary
- make Kubernetes agent pods conform to restricted pod security defaults
- drop privilege escalation and capabilities, and set the expected pod security fields
- keep the change focused on generated pod specs rather than deployment-specific policy

## Validation
- go test ./pkg/runtime
- golangci-lint run --config /Users/mfreeman/src/scion/.golangci.yml --new-from-rev=main ./pkg/runtime


mfreeman451 (commented):
lgtm

PR #76, opened 2026-04-08
Prepare agent home for non-root harness startup

mfreeman451:
## Summary
- precreate harness-specific home directories and the shared `~/.scion` directory in the runtime images
- skip root-only init/git setup when agents already run as a non-root user and expose Codex on the standard shell `PATH`
- avoid hosted permission failures during non-root harness bootstrap, especially around `scion-env` and synchronized home state

## Problem
Hosted non-root agents can fail during startup because the synchronized home layout assumes a few writable directories and init-time behaviors that only work cleanly as root. In practice this showed up as permission failures writing shared Scion state and initializing harness-specific home content.

## Validation
- `go test ./cmd/sciontool/commands`
- hosted Kubernetes proof using `registry.carverauto.dev/scion/scion-codex:home-dirs-a20a5f7b`
  - `Failed to write scion-env file` absent
  - `Error loading configuration: Permission denied` absent

### DIFF
--- a/cmd/sciontool/commands/init.go
+++ b/cmd/sciontool/commands/init.go
@@ -1002,10 +1002,10 @@ func gitCloneWorkspace(uid, gid int, agentHome string) error {
 		}
 	}
 
-	// Ensure the workspace directory is owned by the target user. The
-	// directory may have been created on the host by a root broker process
-	// and bind-mounted into the container as root-owned.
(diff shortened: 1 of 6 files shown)


mfreeman451 (commented):
lgtm

mfreeman451 (commented):
lgtm

PR #78, opened 2026-04-08
Retry dispatched agent updates on conflict

mfreeman451:
## Summary
- retry post-dispatch agent updates when the hub hits a version conflict
- merge broker-returned runtime and container fields into the latest stored row instead of dropping them
- add a focused create-agent regression test for the conflict path

## Validation
- go test ./pkg/hub -run 'TestCreateAgent_(BrokerStatusPreserved|FallbackToProvisioningWhenNoBrokerStatus|RetriesVersionConflictAfterDispatch)$'

### DIFF
--- a/pkg/hub/handlers.go
+++ b/pkg/hub/handlers.go
@@ -811,7 +811,7 @@ func (s *Server) createAgentInGrove(
 				} else if envReqs != nil {
 					// Broker returned 202: needs env gather
 					agent.Phase = string(state.PhaseProvisioning)
-					if err := s.store.UpdateAgent(ctx, agent); err != nil {
+					if err := s.updateAgentAfterDispatch(ctx, agent); err != nil {
 						s.agentLifecycleLog.Warn("Failed to update agent phase for env-gather", "agent_id", agent.ID, "error", err)
 					}
 
@@ -869,7 +869,7 @@ func (s *Server) createAgentInGrove(
 				warnings = append(warnings, "Failed to provision on runtime broker: "+err.Error())
 			} else {
 				agent.Phase = string(state.PhaseCreated)
-				if err := s.store.UpdateAgent(ctx, agent); err != nil {
+				if err := s.updateAgentAfterDispatch(ctx, agent); err != nil {
 					warnings = append(warnings, "Failed to update agent phase: "+err.Error())
 				}
 			}
@@ -919,6 +919,10 @@ func (s *Server) preserveTerminalPhase(ctx context.Context, agent *store.Agent) 
 }
 
 func (s *Server) updateAgentAfterDispatch(ctx context.Context, agent *store.Agent) error {
+	// One retry is intentional here: we only need to recover the common case
+	// where a single concurrent status update bumps StateVersion while dispatch
+	// is in flight. If a second write wins the race too, return the conflict to
+	// the caller rather than spinning in a longer CAS loop inside the request.
 	err := s.store.UpdateAgent(ctx, agent)
 	if err == nil || !errors.Is(err, store.ErrVersionConflict) {
 		return err
--- a/pkg/hub/handlers.go
+++ b/pkg/hub/handlers.go
@@ -831,7 +831,7 @@ func (s *Server) createAgentInGrove(
 					if agent.Phase == string(state.PhaseCreated) {
 						agent.Phase = string(state.PhaseProvisioning)
 					}
-					if err := s.store.UpdateAgent(ctx, agent); err != nil {
+					if err := s.updateAgentAfterDispatch(ctx, agent); err != nil {
 						warnings = append(warnings, "Failed to update agent phase: "+err.Error())
 					}
 				}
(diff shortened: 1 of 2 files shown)


mfreeman451 (commented):
lgtm

ptone:
# PR #78 Review: Retry dispatched agent updates on conflict

## Executive Summary

This PR adds a single-retry mechanism for version-conflict errors when updating an agent after broker dispatch, merging broker-provided fields into the latest stored row. The change is **low-to-medium risk** — the retry logic is straightforward and the merge function correctly guards terminal phases, though there are two consistency gaps worth addressing.

## Critical Issues

### 1. Inconsistent application of `updateAgentAfterDispatch`

**Location:** `pkg/hub/handlers.go:814` and `pkg/hub/handlers.go:872`

Two `UpdateAgent` call sites in `createAgentInGrove` were **not** updated to use the new `updateAgentAfterDispatch`:

- **Line 814** — the env-gather path (`envReqs != nil`, broker returned 202). This path sets `agent.Phase = PhaseProvisioning` and is equally susceptible to the same version-conflict race with concurrent sciontool status reports.
- **Line 872** — the provision-only path (`DispatchAgentProvision`). Same exposure.

If the motivation for this PR is that concurrent status reports bump `StateVersion` during dispatch, these paths share that window and will surface the same bug under load.

**Suggested Fix:** Either convert these two call sites to `updateAgentAfterDispatch` as well, or document explicitly why they are exempt (e.g., because the broker call is synchronous and short enough that a conflict is practically impossible in that path).

### 2. Single retry without loop — silent data loss on double conflict

**Location:** `pkg/hub/handlers.go:920-932` (`updateAgentAfterDispatch`)

The retry performs exactly one re-read + merge + update. If a second concurrent write lands between the `GetAgent` and the second `UpdateAgent`, the update fails and the error propagates as a warning string that is appended to the response but otherwise lost — the broker-provided `Runtime`, `ContainerStatus`, and phase are silently dropped.

In practice this may be rare, but it is worth noting: the pattern is a single retry, not a CAS loop. If this is intentional, a brief code comment would clarify. If not, a bounded retry loop (2-3 iterations) would be more robust:

```go
func (s *Server) updateAgentAfterDispatch(ctx context.Context, agent *store.Agent) error {
    const maxRetries = 3
    for i := 0; i < maxRetries; i++ {
        err := s.store.UpdateAgent(ctx, agent)
        if err == nil || !errors.Is(err, store.ErrVersionConflict) {
            return err
        }
        latest, getErr := s.store.GetAgent(ctx, agent.ID)
        if getErr != nil {
            return getErr
        }
        mergeDispatchedAgent(latest, agent)
        agent = latest
    }
    return s.store.UpdateAgent(ctx, agent)
}
```

## Observations

### `mergeDispatchedAgent` field coverage

The merge function copies: `Template`, `Image`, `Runtime`, `AppliedConfig`, `Message`, `TaskSummary`, `Phase`, `Activity`, `ContainerStatus`, `RuntimeState`.

Fields omitted from the merge but present on `store.Agent` and potentially set during dispatch: `RuntimeBrokerID`, `Detached`, `WebPTYEnabled`, `Labels`, `Annotations`.

If the broker/dispatcher ever sets `RuntimeBrokerID` during `DispatchAgentCreate`, this field would be dropped on conflict retry. Verify this is intentional — it likely is since `RuntimeBrokerID` is set before dispatch, but worth confirming.

### `isTerminalAgentPhase` default case

The `default: return false` is correct and safe but slightly redundant with the explicit non-terminal case list. If new phases are added to `state.Phase` in the future, the default-false behavior means new unknown phases will be treated as non-terminal, which is the safer direction (allows overwrite). This is good defensive design.

### Test quality

The `conflictOnceStore` wrapper is a clean, minimal mock that only overrides `UpdateAgent` and delegates everything else via embedding. The test verifies that broker-set `Runtime` and `ContainerStatus` survive the conflict path. Well-structured.

One minor gap: the test doesn't assert that the `failNextUpdate` flag was actually consumed (i.e., that the conflict path was triggered). Adding `assert.False(t, wrapped.failNextUpdate)` after the request would make the test more precise, though the current assertions implicitly verify correctness.

## Positive Feedback

- The terminal-phase guard in `mergeDispatchedAgent` is a thoughtful safeguard — it prevents the retry from accidentally reverting an error/stopped phase set by a concurrent sciontool report. This aligns well with the existing `preserveTerminalPhase` pattern.
- The separation of `mergeDispatchedAgent` as a pure function makes it easy to unit test independently if needed.
- The PR description includes the exact `go test` command to validate, which is excellent for reviewers.

## Final Verdict

**Needs minor rework.** The core retry + merge logic is sound and the test coverage is solid. The two remaining raw `UpdateAgent` call sites (lines 814 and 872) should either be converted or explicitly justified. The single-retry-vs-loop question is lower priority but worth a clarifying comment. Once addressed, this is ready to merge.

mfreeman451:
Addressed the review points on this branch.

Changes:
- switched the env-gather and provision-only paths over to `updateAgentAfterDispatch(...)` so we are no longer leaving raw `UpdateAgent(...)` call sites behind in the conflict-sensitive path
- kept the retry behavior intentionally narrow and added a comment explaining the single-retry choice
- added focused conflict coverage for the updated paths

Current head: `8573cee7`

PR #92, opened 2026-04-08
Support hub IDs in versioned server config

mfreeman451:
## Summary

- add `hub_id` to `V1ServerHubConfig`
- preserve it when converting between versioned settings and `GlobalConfig`
- document it in the settings schema and cover it with focused config tests

## Problem

Versioned `settings.yaml` currently drops `server.hub.hub_id`, so deployments that rely on versioned server config cannot pin a stable Hub ID through that path.

## Validation

- `go test ./pkg/config -run 'TestConvertV1ServerToGlobalConfig_Basic|TestConvertGlobalToV1ServerConfig_RoundTrip|TestLoadGlobalConfig_FromSettingsYAML'`

 
Refs #91


mfreeman451 (commented):
lgtm

PR #95, opened 2026-04-09
Decode file-backed secrets in Kubernetes runtime

### LINKED ISSUE #94: Kubernetes runtime mounts file secrets as base64 text
## Summary
Hosted Kubernetes agents currently project file-backed secrets with their base64-encoded transport value instead of the decoded file bytes.

## Impact
This breaks any hosted Kubernetes workflow that relies on file secrets containing real file content, such as TLS CA bundles, SSH keys, or JSON credentials. In our validation this showed up as OTLP exporter initialization failing because the mounted CA file contained base64 text instead of PEM data.

## Expected behavior
The Kubernetes runtime should handle file secrets the same way the Docker and Podman runtimes already do:
- decode base64 file-secret payloads before writing them
- fall back to the raw value when the secret was not encoded

## Scope
This is a small runtime-only fix plus regression coverage for the Kubernetes secret creation path.

mfreeman451:
## Summary
- decode file-backed secrets before writing them into fallback Kubernetes Secrets
- preserve raw values when the secret payload is not base64-encoded
- add regression coverage for decoded and raw file-secret content

## Validation
- go test ./pkg/runtime -run 'Test(CreateAgentSecret|BuildPod_FallbackSecrets_File|CreateAgentSecret_AlreadyExists)$'
- golangci-lint run --config /Users/mfreeman/src/scion/.golangci.yml --new-from-rev=origin/main ./pkg/runtime

Closes #94

### DIFF
--- a/pkg/runtime/k8s_runtime.go
+++ b/pkg/runtime/k8s_runtime.go
@@ -380,7 +380,11 @@ func (r *KubernetesRuntime) createAgentSecret(ctx context.Context, namespace, ag
 		case "environment":
 			data[s.Name] = []byte(s.Value)
 		case "file":
-			data[s.Name] = []byte(s.Value)
+			decoded, err := base64.StdEncoding.DecodeString(s.Value)
+			if err != nil {
+				decoded = []byte(s.Value)
+			}
+			data[s.Name] = decoded
 		case "variable":
 			varSecrets[s.Target] = s.Value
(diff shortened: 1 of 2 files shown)


mfreeman451 (commented):
lgtm

mfreeman451 (commented):
lgtm

PR #99, opened 2026-04-09
fix: chown home directory before privilege drop in sciontool

### LINKED ISSUE #97: sciontool: home directory files owned by root after pre-start hooks, child can't write
## Problem

When `sciontool init` runs as root (PID 1) in a container, it executes pre-start hooks that create files in the user's home directory:

- `.claude/` directory (CLAUDE.md, settings.json, skills/)
- `.claude.json`
- `agent-info.json`
- `.scion/scion-env`
- `.bashrc`, `.zshrc`, `.tmux.conf`, `.gitconfig`, `.gemini/`

These files are created as root. sciontool then drops privileges to the `scion` user (UID 1000, via `SCION_HOST_UID`) before launching the child process. The child process (e.g. Claude Code) cannot write to its own config directory:

```
$ ls -la /home/scion/
drwx------ 5 root  root  4096 .
drwxr-xr-x 4 root  root  4096 .claude
-rw-r--r-- 1 root  root   977 .claude.json
drwx------ 2 scion scion 4096 .scion    # only .scion is correctly owned
```

Claude Code starts but produces no output — it silently fails because it can't write auth state or config to `~/.claude/`.

## Reproduction

1. Use rootful Podman (or any container runtime where the image starts as `USER root`)
2. Start an agent with a custom image
3. Observe that home directory files are root-owned
4. The harness (Claude, Gemini, etc.) fails to write config

## Expected behaviour

After pre-start hooks run and before the child process launches, sciontool should `chown -R` the home directory to the target UID:GID (`SCION_HOST_UID:SCION_HOST_GID`).

The fix belongs in `pkg/sciontool/supervisor/supervisor.go`, in the `Run` method, between the env setup and `s.cmd.Start()`.

## Environment

- Podman 4.9.3
- SCION commit a9fca6cb
- Linux 6.8.0
- Custom image with `USER root` (required for sciontool to chown mounted volumes)

meatballs:
## Summary

- sciontool now recursively chowns the home directory to the target UID:GID before launching the child process
- Pre-start hooks run as root and create files (`.claude/`, `agent-info.json`, `.scion/scion-env`) that the child process needs to write to
- Without this, harnesses like Claude Code fail silently on startup because they can't write auth state or config

Fixes #97

## Test plan

- [x] Existing supervisor tests pass
- [x] Verified manually: after fix, `/home/scion/` contents are owned by scion:scion
- [x] Claude Code starts successfully and can write to `~/.claude/`

### DIFF
--- a/pkg/sciontool/supervisor/supervisor.go
+++ b/pkg/sciontool/supervisor/supervisor.go
@@ -125,8 +125,9 @@ func (s *Supervisor) Run(ctx context.Context, args []string) (int, error) {
 		err := chownRecursive(home, s.config.UID, s.config.GID)
 		if err != nil {
 			log.Error("Failed to chown home directory %s: %v", home, err)
+		} else {
+			log.Debug("Chowned %s to %d:%d", home, s.config.UID, s.config.GID)
 		}
-		log.Debug("Chowned %s to %d:%d", home, s.config.UID, s.config.GID)
 	}
 
 	// Apply SCION_EXTRA_PATH: prepend its value to PATH, then remove it from env.
--- a/pkg/sciontool/supervisor/supervisor.go
+++ b/pkg/sciontool/supervisor/supervisor.go
@@ -12,6 +12,7 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"sync"
 	"syscall"
 	"time"
@@ -113,6 +114,19 @@ func (s *Supervisor) Run(ctx context.Context, args []string) (int, error) {
 		env = setEnvVar(env, "LOGNAME", s.config.Username)
 		s.cmd.Env = env
 		log.Debug("Child env: HOME=%s, USER=%s, LOGNAME=%s", home, s.config.Username, s.config.Username)
+	}
+
+	// Fix ownership of the home directory when dropping privileges.
+	// Pre-start hooks run as root and may create files (e.g. .claude/, agent-info.json)
+	// that the child process needs to write to. Without this, the child (running as
+	// UID/GID from the credential drop) gets permission denied on its own home.
+	if s.config.UID > 0 && s.config.GID > 0 && s.config.Username != "" {
+		home := "/home/" + s.config.Username
+		if err := chownRecursive(home, s.config.UID, s.config.GID); err != nil {
+			log.Error("Failed to chown home directory %s: %v", home, err)
+		} else {
+			log.Debug("Chowned %s to %d:%d", home, s.config.UID, s.config.GID)
+		}
 	}
 
 	// Apply SCION_EXTRA_PATH: prepend its value to PATH, then remove it from env.
@@ -297,4 +311,14 @@ func removeEnvVar(env []string, key string) []string {
 		result = append(result, e)
 	}
 	return result
+}
+
+// chownRecursive changes ownership of a directory and all its contents.
+func chownRecursive(root string, uid, gid int) error {
+	return filepath.Walk(root, func(path string, info os.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+		return os.Lchown(path, uid, gid)
+	})
 }
--- a/pkg/sciontool/supervisor/supervisor.go
+++ b/pkg/sciontool/supervisor/supervisor.go
@@ -122,11 +122,11 @@ func (s *Supervisor) Run(ctx context.Context, args []string) (int, error) {
 	// UID/GID from the credential drop) gets permission denied on its own home.
 	if s.config.UID > 0 && s.config.GID > 0 && s.config.Username != "" {
 		home := "/home/" + s.config.Username
-		if err := chownRecursive(home, s.config.UID, s.config.GID); err != nil {
+		err := chownRecursive(home, s.config.UID, s.config.GID)
+		if err != nil {
 			log.Error("Failed to chown home directory %s: %v", home, err)
-		} else {
-			log.Debug("Chowned %s to %d:%d", home, s.config.UID, s.config.GID)
 		}
+		log.Debug("Chowned %s to %d:%d", home, s.config.UID, s.config.GID)
 	}
 
 	// Apply SCION_EXTRA_PATH: prepend its value to PATH, then remove it from env.


pkg/sciontool/supervisor/supervisor.go
```diff
@@ -115,6 +116,19 @@ func (s *Supervisor) Run(ctx context.Context, args []string) (int, error) {
...
 
+	// Fix ownership of the home directory when dropping privileges.
+	// Pre-start hooks run as root and may create files (e.g. .claude/, agent-info.json)
+	// that the child process needs to write to. Without this, the child (running as
+	// UID/GID from the credential drop) gets permission denied on its own home.
+	if s.config.UID > 0 && s.config.GID > 0 && s.config.Username != "" {
+		home := "/home/" + s.config.Username
+		if err := chownRecursive(home, s.config.UID, s.config.GID); err != nil {
+			log.Error("Failed to chown home directory %s: %v", home, err)
+		} else {
```
mfreeman451: consider using guard-clauses here

ptone:
# PR #99 Review: fix: chown home directory before privilege drop in sciontool

**PR:** https://github.com/GoogleCloudPlatform/scion/pull/99
**Branch:** `fix/sciontool-chown-home` -> `main`
**Reviewed:** 2026-04-11

---

## Executive Summary

This PR adds a recursive chown of the scion user's home directory inside `supervisor.Run()`, immediately before the child process launches. The change addresses a real gap where pre-start hooks (running as root) create files like `.claude/` and `agent-info.json` that the privilege-dropped child process cannot write to. **The fix is NOT already present on `main`** — while the codebase has several ownership-fixing mechanisms (Dockerfile build-time chown, `directSetUID()` chown during UID remapping, K8s runtime chown after host sync), none of them run *after* the pre-start lifecycle hooks and *before* the supervisor starts the child process.

**Risk Level: Low-Medium.** The change is small and well-scoped, but has a few issues worth addressing.

---

## Is This Already Fixed?

**No.** The existing ownership mechanisms on `main` are:

| Mechanism | When it runs | Gap |
|---|---|---|
| Dockerfile `chown -R scion:scion /home/scion` | Build time | Pre-start hooks run later at runtime and create root-owned files |
| `directSetUID()` in init.go | During UID remapping (before hooks) | Runs *before* `lifecycleManager.RunPreStart()`, so hook-created files are missed |
| K8s runtime chown | After host-first sync | Only applies to K8s host-sync path, not all deployment modes |
| Individual file chowns (log, token, env) | Various points in init | Doesn't cover `.claude/` or `agent-info.json` created by status handler |

The pre-start hooks at init.go:229 create `agent-info.json` (via `StatusHandler`) as root. The supervisor doesn't launch until line ~338+. Nothing between those points chowns the home directory recursively. **This PR fills a real gap.**

---

## Critical Issues

### 1. Error is logged but execution continues to log the success message

**File:** `pkg/sciontool/supervisor/supervisor.go`, lines 126-129

```go
err := chownRecursive(home, s.config.UID, s.config.GID)
if err != nil {
    log.Error("Failed to chown home directory %s: %v", home, err)
}
log.Debug("Chowned %s to %d:%d", home, s.config.UID, s.config.GID)
```

The `log.Debug("Chowned ...")` line runs unconditionally, even when `chownRecursive` fails. This is misleading in logs. While not a functional bug (the PR intentionally treats this as non-fatal, matching the pattern at init.go:230-231), the success log should be inside an `else` block.

**Suggested Fix:**
```go
err := chownRecursive(home, s.config.UID, s.config.GID)
if err != nil {
    log.Error("Failed to chown home directory %s: %v", home, err)
} else {
    log.Debug("Chowned %s to %d:%d", home, s.config.UID, s.config.GID)
}
```

### 2. Hardcoded `/home/` path duplicates and diverges from existing logic

**File:** `pkg/sciontool/supervisor/supervisor.go`, line 124

```go
home := "/home/" + s.config.Username
```

This duplicates the same pattern from line 110, but introduces a subtle divergence risk. The `home` variable at line 110 is scoped inside the `if` block above and isn't available here. If the home directory resolution logic ever changes (e.g., to support non-standard home paths), this hardcoded path won't be updated.

**Suggested Fix:** Consider adding `HomeDir` to `Config` so `init.go` passes the resolved `agentHome` rather than reconstructing it:

```go
// In Config:
HomeDir string

// In init.go where config is created (line 331):
config := supervisor.Config{
    // ...
    HomeDir: agentHome,
}

// In supervisor.go:
if s.config.UID > 0 && s.config.GID > 0 && s.config.HomeDir != "" {
    err := chownRecursive(s.config.HomeDir, s.config.UID, s.config.GID)
    // ...
}
```

This is a design improvement, not a blocker.

---

## Observations

### 3. Performance: Recursive walk on every supervisor start

`filepath.Walk` + `os.Lchown` on every file in `/home/scion/` runs synchronously before the child starts. For a fresh container with a small home directory (skeleton files + `.claude/` + `agent-info.json`), this is negligible. However, if the home directory grows large (e.g., after git clone populates a workspace under the home tree, or large config caches), this could add noticeable startup latency.

The current scope is safe since `/home/scion` is typically small at this point in the init flow (git clone writes to `/workspace`, not `/home/scion`).

### 4. `filepath.Walk` follows symlinks for the root but not for children

`filepath.Walk` will follow a symlink if `root` itself is a symlink, but uses `os.Lstat` for children (so it won't follow symlinks to directories outside the tree). The use of `os.Lchown` (which doesn't follow symlinks) is correct and avoids accidentally chowning files outside the home directory via symlinks. Good choice.

### 5. Condition check is slightly more restrictive than needed

Line 123: `if s.config.UID > 0 && s.config.GID > 0 && s.config.Username != ""`

This is more restrictive than the privilege-drop condition (line 96: `if s.config.UID > 0 && s.config.GID > 0`). The extra `Username != ""` check is reasonable as a guard since the home path depends on username, but note that in practice `Username` is always `"scion"` when UID/GID are set (see init.go:335). This is fine.

---

## Positive Feedback

- **Good use of `os.Lchown`** instead of `os.Chown` — prevents symlink traversal attacks where a malicious symlink in the home directory could cause ownership changes to files outside the tree.
- **Non-fatal error handling** is consistent with the project pattern (init.go:230-231 also logs and continues when pre-start hooks fail).
- **Clear, well-written comments** explaining *why* the chown is needed.
- The `chownRecursive` function is clean and minimal.

---

## Final Verdict

**Approve with minor fixes.** The PR addresses a real, unpatched gap in the ownership chain. The fix is correctly placed (after pre-start hooks, before child exec) and uses secure APIs (`Lchown`). The two issues above (misleading success log, hardcoded path) are minor — the log issue should be fixed before merge; the path duplication is a nice-to-have improvement.

meatballs:
Thanks for the review @ptone — addressed the success-log issue in `81bb1932` by moving `log.Debug("Chowned ...")` into an `else` branch. Left the hardcoded `/home/` path as-is for now; happy to thread `HomeDir` through `supervisor.Config` as a follow-up if you'd prefer it bundled here.

PR #102, opened 2026-04-09
fix: proxy local storage uploads through hub HTTP endpoint

### LINKED ISSUE #100: Template sync fails with 'mkdir /home/scion: permission denied' on hub storage
## Problem

`scion template sync` fails when uploading template files to the hub, even though the target directory exists and is writable:

```
$ scion template sync empiria-agent
Scanning template files in /home/owen/.scion/templates/empiria-agent...
Found 4 files
Creating template 'empiria-agent' in Hub...
Template created with ID: e300c83f-...
Requesting upload URLs for 4 file(s)...
Uploading 4 file(s)...

Error: failed to upload agents.md: failed to create directory
  /home/scion/.scion/storage/local/templates/groves/<grove-id>/empiria-agent:
  mkdir /home/scion: permission denied
```

The hub process (UID 1000, user `scion`) can write to the target directory — verified by `podman exec scion-hub touch <path>`. But the code's `os.MkdirAll` call tries to create the full path from `/home/scion` upward, and fails because it cannot recreate the mount point.

This appears to be a path resolution issue where `MkdirAll` is called with an absolute path but something in the resolution strips or re-derives the base, causing it to attempt `mkdir` on directories it shouldn't need to create.

## Impact

Without working template sync, custom templates (with volumes, env vars, agent instructions) cannot be pushed to the hub. Hub-dispatched agents fall back to the `default` template, losing all custom configuration. This effectively means custom agent setups only work in local mode (`--no-hub`).

## Reproduction

1. Deploy a hub with a bind mount at `/home/scion` (or `/home/scion/.scion`)
2. Register a grove and a local template
3. Run `scion template sync <template-name>`
4. Observe the `mkdir /home/scion: permission denied` error

The error occurs regardless of whether the volume is mounted at `/home/scion` or `/home/scion/.scion`. The target directory and all parent directories within the mount exist and are writable.

## Workaround

None currently. Manually copying files to the hub's storage directory doesn't work because the hub's database has no record of the files (the template record is created but the file upload fails).

## Environment

- Hub deployed via Podman (rootful, on ARM64)
- Hub image built from SCION commit a9fca6cb
- Client: SCION a9fca6cb, Linux 6.8.0

meatballs:
## Summary

Fixes #100 — `scion template sync` fails with `mkdir /home/scion: permission denied` when the hub uses local filesystem storage.

**Root cause:** Local storage generates `file://` signed URLs containing server-side absolute paths (e.g. `file:///home/scion/.scion/storage/local/templates/...`). The client receives these URLs and attempts to write/read at those paths on the *local* machine, which fails because the hub's filesystem isn't accessible remotely.

**Fix:** When the hub's storage backend is local, rewrite `file://` signed URLs to HTTP proxy URLs pointing to the hub's own template files endpoint. Both uploads (PUT) and downloads (GET) are proxied. The client's existing HTTP code handles these transparently.

### Commit 1: Upload proxy

- `storage_helpers.go`: Add `requestBaseURL()` (derives external URL from request headers, supports reverse proxies) and `rewriteLocalUploadURLs()` (converts `file://` to HTTP PUT URLs with auth headers)
- `template_file_handlers.go`: Add raw binary PUT support via `handleTemplateFileWriteRaw` — detects non-JSON Content-Type and writes body directly to storage
- `template_handlers.go`: Wire up URL rewriting in template create and upload handlers

### Commit 2: Download proxy + authenticated transfer clients

- `storage_helpers.go`: Add `rewriteLocalDownloadURLs()` — converts `file://` to HTTP GET URLs with `?raw=1` param
- `template_file_handlers.go`: Add raw binary streaming to `handleTemplateFileRead` — when `?raw=1` or `Accept: application/octet-stream`, streams file content directly instead of JSON wrapper (bypasses 1MB limit)
- `template_handlers.go`: Wire up download URL rewriting
- `apiclient/transport.go`: Add `AuthenticatedHTTPClient()` — returns an `*http.Client` with an auth-injecting RoundTripper, so proxy URLs carry proper auth
- `hubclient/{templates,workspace,harness_configs}.go`: Use `AuthenticatedHTTPClient()` for transfer clients

### No client-side changes required

The client already handles HTTP upload/download URLs (used with GCS storage). The fix is entirely server-side — the hub returns different URLs, and existing client code works unchanged.

## Test plan

- [x] `go build ./...` — compiles cleanly
- [x] `go vet` — no warnings
- [x] Tested with self-hosted hub (Podman, ARM64) + remote client — `scion template sync` uploads all files successfully
- [x] Template download by broker works — template hydration succeeds with correct content hashes
- [ ] Verify GCS storage path is unaffected (no `file://` URLs generated, rewrite is a no-op)

### DIFF
--- a/pkg/hub/storage_helpers.go
+++ b/pkg/hub/storage_helpers.go
@@ -138,8 +138,9 @@ func generateDownloadURLs(ctx context.Context, stor storage.Storage, basePath st
 //
 // with method PUT. The client's authenticated HTTP transport handles auth.
 // requestBaseURL derives the external base URL from the incoming HTTP request.
-// It uses X-Forwarded-Proto and X-Forwarded-Host if set (reverse proxy), falling
-// back to the request's TLS state and Host header.
+// It trusts X-Forwarded-Proto and X-Forwarded-Host when present, which assumes
+// the hub is behind a trusted reverse proxy that overwrites those headers.
+// Otherwise it falls back to the request's TLS state and Host header.
 func requestBaseURL(r *http.Request) string {
 	host := r.Header.Get("X-Forwarded-Host")
 	if host == "" {
@@ -155,7 +156,7 @@ func requestBaseURL(r *http.Request) string {
 	return "http://" + host
 }
 
-func rewriteLocalUploadURLs(urls []UploadURLInfo, hubEndpoint, resourceType, resourceID, authHeader string) []UploadURLInfo {
+func rewriteLocalUploadURLs(urls []UploadURLInfo, hubEndpoint, resourceType, resourceID string) []UploadURLInfo {
 	if hubEndpoint == "" {
 		return urls
 	}
@@ -167,9 +168,6 @@ func rewriteLocalUploadURLs(urls []UploadURLInfo, hubEndpoint, resourceType, res
 			urls[i].Headers = map[string]string{
 				"Content-Type": "application/octet-stream",
 			}
-			if authHeader != "" {
-				urls[i].Headers["Authorization"] = authHeader
-			}
 		}
 	}
 	return urls
@@ -179,7 +177,7 @@ func rewriteLocalUploadURLs(urls []UploadURLInfo, hubEndpoint, resourceType, res
 // the hub's own file read endpoint for downloads. Same rationale as
 // rewriteLocalUploadURLs — file:// URLs reference server-side paths that are
 // inaccessible from remote clients.
-func rewriteLocalDownloadURLs(urls []DownloadURLInfo, hubEndpoint, resourceType, resourceID, authHeader string) []DownloadURLInfo {
+func rewriteLocalDownloadURLs(urls []DownloadURLInfo, hubEndpoint, resourceType, resourceID string) []DownloadURLInfo {
 	if hubEndpoint == "" {
 		return urls
 	}
--- a/pkg/hub/storage_helpers.go
+++ b/pkg/hub/storage_helpers.go
@@ -16,6 +16,8 @@ package hub
 
 import (
 	"context"
+	"net/http"
+	"strings"
 	"time"
 
 	"github.com/GoogleCloudPlatform/scion/pkg/storage"
@@ -122,4 +124,51 @@ func generateDownloadURLs(ctx context.Context, stor storage.Storage, basePath st
 	}
 
 	return downloadURLs, manifestURL, expires, nil
+}
+
+// rewriteLocalUploadURLs rewrites file:// URLs to HTTP proxy URLs pointing to
+// the hub's own file upload endpoint. This is necessary because local storage
+// generates file:// signed URLs that reference server-side paths, which are not
+// accessible when the client is on a different machine.
+//
+// For each URL with a file:// scheme, it is replaced with:
+//
+//	<hubEndpoint>/api/v1/<resourceType>/<resourceID>/files/<filePath>
+//
+// with method PUT. The client's authenticated HTTP transport handles auth.
+// requestBaseURL derives the external base URL from the incoming HTTP request.
+// It uses X-Forwarded-Proto and X-Forwarded-Host if set (reverse proxy), falling
+// back to the request's TLS state and Host header.
+func requestBaseURL(r *http.Request) string {
+	scheme := "http"
+	if proto := r.Header.Get("X-Forwarded-Proto"); proto != "" {
+		scheme = proto
+	} else if r.TLS != nil {
+		scheme = "https"
+	}
+	host := r.Header.Get("X-Forwarded-Host")
+	if host == "" {
+		host = r.Host
+	}
+	return scheme + "://" + host
+}
+
+func rewriteLocalUploadURLs(urls []UploadURLInfo, hubEndpoint, resourceType, resourceID, authHeader string) []UploadURLInfo {
+	if hubEndpoint == "" {
+		return urls
+	}
+	hubEndpoint = strings.TrimRight(hubEndpoint, "/")
+	for i := range urls {
+		if strings.HasPrefix(urls[i].URL, "file://") {
+			urls[i].URL = hubEndpoint + "/api/v1/" + resourceType + "/" + resourceID + "/files/" + urls[i].Path
+			urls[i].Method = "PUT"
+			urls[i].Headers = map[string]string{
+				"Content-Type": "application/octet-stream",
+			}
+			if authHeader != "" {
+				urls[i].Headers["Authorization"] = authHeader
+			}
+		}
+	}
+	return urls
 }
--- a/pkg/hub/storage_helpers.go
+++ b/pkg/hub/storage_helpers.go
@@ -16,6 +16,7 @@ package hub
 
 import (
 	"context"
+	"fmt"
 	"net/http"
 	"strings"
 	"time"
@@ -160,8 +161,8 @@ func rewriteLocalUploadURLs(urls []UploadURLInfo, hubEndpoint, resourceType, res
 	hubEndpoint = strings.TrimRight(hubEndpoint, "/")
 	for i := range urls {
 		if strings.HasPrefix(urls[i].URL, "file://") {
-			urls[i].URL = hubEndpoint + "/api/v1/" + resourceType + "/" + resourceID + "/files/" + urls[i].Path
-			urls[i].Method = "PUT"
+			urls[i].URL = fmt.Sprintf("%s/api/v1/%s/%s/files/%s", hubEndpoint, resourceType, resourceID, urls[i].Path)
+			urls[i].Method = http.MethodPut
 			urls[i].Headers = map[string]string{
 				"Content-Type": "application/octet-stream",
 			}
--- a/pkg/hub/storage_helpers.go
+++ b/pkg/hub/storage_helpers.go
@@ -141,17 +141,18 @@ func generateDownloadURLs(ctx context.Context, stor storage.Storage, basePath st
 // It uses X-Forwarded-Proto and X-Forwarded-Host if set (reverse proxy), falling
 // back to the request's TLS state and Host header.
 func requestBaseURL(r *http.Request) string {
-	scheme := "http"
-	if proto := r.Header.Get("X-Forwarded-Proto"); proto != "" {
-		scheme = proto
-	} else if r.TLS != nil {
-		scheme = "https"
-	}
 	host := r.Header.Get("X-Forwarded-Host")
 	if host == "" {
 		host = r.Host
 	}
-	return scheme + "://" + host
+
+	if proto := r.Header.Get("X-Forwarded-Proto"); proto != "" {
+		return proto + "://" + host
+	}
+	if r.TLS != nil {
+		return "https://" + host
+	}
+	return "http://" + host
 }
 
 func rewriteLocalUploadURLs(urls []UploadURLInfo, hubEndpoint, resourceType, resourceID, authHeader string) []UploadURLInfo {
--- a/pkg/hub/storage_helpers.go
+++ b/pkg/hub/storage_helpers.go
@@ -172,3 +172,20 @@ func rewriteLocalUploadURLs(urls []UploadURLInfo, hubEndpoint, resourceType, res
 	}
 	return urls
 }
+
+// rewriteLocalDownloadURLs rewrites file:// URLs to HTTP proxy URLs pointing to
+// the hub's own file read endpoint for downloads. Same rationale as
+// rewriteLocalUploadURLs — file:// URLs reference server-side paths that are
+// inaccessible from remote clients.
+func rewriteLocalDownloadURLs(urls []DownloadURLInfo, hubEndpoint, resourceType, resourceID, authHeader string) []DownloadURLInfo {
+	if hubEndpoint == "" {
+		return urls
+	}
+	hubEndpoint = strings.TrimRight(hubEndpoint, "/")
+	for i := range urls {
+		if strings.HasPrefix(urls[i].URL, "file://") {
+			urls[i].URL = hubEndpoint + "/api/v1/" + resourceType + "/" + resourceID + "/files/" + urls[i].Path + "?raw=1"
+		}
+	}
+	return urls
+}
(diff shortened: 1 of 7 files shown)


pkg/hub/storage_helpers.go
```diff
@@ -123,3 +125,67 @@ func generateDownloadURLs(ctx context.Context, stor storage.Storage, basePath st
...
+
+func rewriteLocalUploadURLs(urls []UploadURLInfo, hubEndpoint, resourceType, resourceID, authHeader string) []UploadURLInfo {
+	if hubEndpoint == "" {
+		return urls
+	}
+	hubEndpoint = strings.TrimRight(hubEndpoint, "/")
+	for i := range urls {
+		if strings.HasPrefix(urls[i].URL, "file://") {
+			urls[i].URL = hubEndpoint + "/api/v1/" + resourceType + "/" + resourceID + "/files/" + urls[i].Path
+			urls[i].Method = "PUT"
```
mfreeman451: should use `http.MethodPut` here instead of plaintext "PUT"

pkg/hub/storage_helpers.go
```diff
@@ -123,3 +125,67 @@ func generateDownloadURLs(ctx context.Context, stor storage.Storage, basePath st
...
+	}
+	return scheme + "://" + host
+}
+
+func rewriteLocalUploadURLs(urls []UploadURLInfo, hubEndpoint, resourceType, resourceID, authHeader string) []UploadURLInfo {
+	if hubEndpoint == "" {
+		return urls
+	}
+	hubEndpoint = strings.TrimRight(hubEndpoint, "/")
+	for i := range urls {
```
mfreeman451, line 164: This might be a better approach:

```
type UploadURL struct {
    URL    string
    Method string
    Path   string
    // ...
}

for i := range urls {
    if strings.HasPrefix(urls[i].URL, "file://") {
        urls[i].URL = fmt.Sprintf("%s/api/v1/%s/%s/files/%s",
            strings.TrimRight(hubEndpoint, "/"),
            resourceType,
            resourceID,
            urls[i].Path,
        )
        urls[i].Method = http.MethodPut
    }
}
```

pkg/hub/template_file_handlers.go
```diff
@@ -158,19 +162,39 @@ func (s *Server) handleTemplateFileRead(w http.ResponseWriter, r *http.Request,
...
 	stor := s.GetStorage()
 	if stor == nil {
 		RuntimeError(w, "Storage not configured")
 		return
 	}
 
+	// Raw binary download for local storage proxy flow
+	if r.URL.Query().Get("raw") != "" || strings.Contains(r.Header.Get("Accept"), "application/octet-stream") {
+		objectPath := template.StoragePath + "/" + filePath
+		reader, _, err := stor.Download(ctx, objectPath)
```
mfreeman451, line 179: prefer errors.Is here, `if errors.Is(err, storage.ErrNotFound) {`

pkg/hub/template_file_handlers.go
```diff
@@ -158,19 +162,39 @@ func (s *Server) handleTemplateFileRead(w http.ResponseWriter, r *http.Request,
...
+				return
+			}
+			RuntimeError(w, "Failed to read file from storage")
+			return
+		}
+		defer reader.Close()
+
+		w.Header().Set("Content-Type", "application/octet-stream")
+		w.WriteHeader(http.StatusOK)
+		io.Copy(w, reader)
```
mfreeman451: should log error here, 

```if _, err := io.Copy(w, reader); err != nil {
        // Log the error but don't send another response (headers already sent)
        log.Printf("Error streaming file %s: %v", objectPath, err)
    }
```

pkg/hub/template_file_handlers.go
```diff
@@ -158,19 +162,39 @@ func (s *Server) handleTemplateFileRead(w http.ResponseWriter, r *http.Request,
...
+			if err == storage.ErrNotFound {
+				NotFound(w, "Template file")
+				return
+			}
+			RuntimeError(w, "Failed to read file from storage")
+			return
+		}
+		defer reader.Close()
+
+		w.Header().Set("Content-Type", "application/octet-stream")
```
mfreeman451, line 193: nit: add 
`w.Header().Set("Content-Disposition", `attachment; filename="`+filePath+`"`)

below the Content-type line

mfreeman451 (commented):
comments inline

pkg/hub/template_file_handlers.go
mfreeman451: consider using `http.ServeContent(w, r, filePath, meta.ModTime, seeker)` instead here

pkg/hub/storage_helpers.go
```diff
@@ -123,3 +126,67 @@ func generateDownloadURLs(ctx context.Context, stor storage.Storage, basePath st
...
+//
+//	<hubEndpoint>/api/v1/<resourceType>/<resourceID>/files/<filePath>
+//
+// with method PUT. The client's authenticated HTTP transport handles auth.
+// requestBaseURL derives the external base URL from the incoming HTTP request.
+// It uses X-Forwarded-Proto and X-Forwarded-Host if set (reverse proxy), falling
+// back to the request's TLS state and Host header.
+func requestBaseURL(r *http.Request) string {
+	scheme := "http"
+	if proto := r.Header.Get("X-Forwarded-Proto"); proto != "" {
```
mfreeman451: consider using guard clauses here

ptone:
# PR-102 Review: fix: proxy local storage uploads through hub HTTP endpoint

**Reviewer:** Scion Code Review Agent
**Date:** 2026-04-11
**PR:** #102 — Fix `scion template sync` with local filesystem storage

---

## Executive Summary

This PR solves a real architectural gap where local-storage `file://` signed URLs are unusable by remote clients, by proxying uploads/downloads through the hub's HTTP API. The change is **moderate risk**: the approach is sound and well-structured, but there are two security issues (one critical, one medium) and a Go API contract violation that should be addressed before merge.

---

## Critical Issues

### 1. Header Injection via `Content-Disposition` (Security — High)

**File:** `pkg/hub/template_file_handlers.go` (new raw download path)

```go
w.Header().Set("Content-Disposition", `attachment; filename="`+filePath+`"`)
```

`filePath` is derived from the URL path (`strings.TrimPrefix(action, "files/")` in `template_handlers.go:319`) and is only validated against the template manifest (existence check). A file path containing `"` or CRLF characters could break the header value or inject additional headers.

While the manifest check limits exploitability (the path must match an existing file), the manifest itself is user-created — a malicious user could register a file with a crafted name like `foo"\r\nX-Injected: evil`.

**Suggested Fix:**
```go
import "mime"

// Sanitize filename for Content-Disposition
safeName := filepath.Base(filePath)
w.Header().Set("Content-Disposition", mime.FormatMediaType("attachment", map[string]string{"filename": safeName}))
```

### 2. `authRoundTripper.RoundTrip` Mutates the Request (Correctness — High)

**File:** `pkg/apiclient/transport.go`

```go
func (rt *authRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
	if rt.ua != "" && req.Header.Get("User-Agent") == "" {
		req.Header.Set("User-Agent", rt.ua)       // ← mutates original request
	}
	if rt.auth != nil {
		if err := rt.auth.ApplyAuth(req); err != nil {  // ← mutates original request
```

The [`http.RoundTripper` contract](https://pkg.go.dev/net/http#RoundTripper) states: "RoundTrip should not modify the request, except for consuming and closing the Body." Mutating headers on the original request is a violation that can cause data races when requests are retried or used concurrently.

**Suggested Fix:**
```go
func (rt *authRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
	// Clone the request to avoid mutating the original per RoundTripper contract.
	clone := req.Clone(req.Context())
	if rt.ua != "" && clone.Header.Get("User-Agent") == "" {
		clone.Header.Set("User-Agent", rt.ua)
	}
	if rt.auth != nil {
		if err := rt.auth.ApplyAuth(clone); err != nil {
			return nil, fmt.Errorf("failed to apply auth: %w", err)
		}
	}
	base := rt.base
	if base == nil {
		base = http.DefaultTransport
	}
	return base.RoundTrip(clone)
}
```

### 3. Authorization Header Leaked to External Hosts (Security — Medium)

**File:** `pkg/hub/storage_helpers.go` — `rewriteLocalUploadURLs`

```go
if authHeader != "" {
    urls[i].Headers["Authorization"] = authHeader
}
```

The `Authorization` header from the incoming request is embedded directly into the `UploadURLInfo` response sent to the client. For local storage this works because the URLs point back to the hub, but:

1. The auth header is now serialized into a JSON response and sent over the wire — expanding its exposure surface.
2. If any code path accidentally calls `rewriteLocalUploadURLs` without the `ProviderLocal` guard, or if the URLs are logged/cached, the credential leaks.

This is partially mitigated by the `ProviderLocal` check at the call sites. However, consider whether this is even necessary given that `AuthenticatedHTTPClient()` is now used on the client side (which already injects auth). The auth header in the URL response may be **redundant** and could be removed entirely.

**Suggested investigation:** Verify whether the transfer client on the client side uses the `Headers` from `UploadURLInfo` when making PUT requests. If `AuthenticatedHTTPClient()` already injects auth, remove the `Authorization` header from the rewritten URLs to minimize credential exposure.

---

## Observations

### 4. `requestBaseURL` Trusts Proxy Headers Without Validation (Low)

**File:** `pkg/hub/storage_helpers.go`

```go
func requestBaseURL(r *http.Request) string {
	host := r.Header.Get("X-Forwarded-Host")
	if host == "" {
		host = r.Host
	}
	if proto := r.Header.Get("X-Forwarded-Proto"); proto != "" {
		return proto + "://" + host
	}
```

`X-Forwarded-Host` and `X-Forwarded-Proto` are trivially spoofable by any client. A malicious client could set `X-Forwarded-Host: evil.com` and receive rewritten URLs pointing to `http://evil.com/api/v1/templates/...`, potentially redirecting subsequent uploads to an attacker-controlled server.

This is a known pattern in Go HTTP servers and is acceptable **only if** the hub is deployed behind a trusted reverse proxy that strips/overwrites these headers. Consider documenting this assumption or adding an allowlist.

### 5. Missing `Content-Length` on Raw Download

**File:** `pkg/hub/template_file_handlers.go`

The raw download path streams via `io.Copy` without setting `Content-Length`. The `found.Size` is available from the manifest. Setting it would enable clients to show download progress and detect truncation:

```go
w.Header().Set("Content-Length", strconv.FormatInt(found.Size, 10))
```

### 6. `handleTemplateFileWriteRaw` Reads Entire File Into Memory

**File:** `pkg/hub/template_file_handlers.go`

```go
data, err := io.ReadAll(io.LimitReader(r.Body, maxUploadFileSize+1))
```

For files up to 50MB, this allocates the full content in memory. The SHA256 hash computation also requires the full content, so streaming isn't straightforward here. This is acceptable for the current 50MB limit but worth noting if the limit increases.

### 7. No `r.Body.Close()` in Raw Upload Path

The raw upload path reads `r.Body` via `io.ReadAll` but doesn't explicitly close it. While Go's HTTP server closes the body after the handler returns, explicitly closing after `ReadAll` would release resources sooner for large uploads.

### 8. `AuthenticatedHTTPClient` Doesn't Copy All `http.Client` Fields

**File:** `pkg/apiclient/transport.go`

```go
return &http.Client{
    Timeout:   base.Timeout,
    Transport: &authRoundTripper{...},
}
```

`CheckRedirect` and `Jar` from the base client are not copied. This is likely fine for the current use case (simple PUT/GET to hub URLs), but worth noting if redirect behavior matters in the future.

---

## Positive Feedback

- **Clean architectural approach:** Rewriting URLs server-side so the client needs zero changes is elegant and minimizes blast radius.
- **Consistent application:** The `ProviderLocal` check + URL rewrite is applied uniformly across create, upload, and download paths.
- **Good use of existing patterns:** The raw binary handlers mirror the existing JSON handlers' structure (manifest update, content hash recomputation, storage upload).
- **Proper `LimitReader` usage:** The `maxUploadFileSize+1` pattern to detect oversized uploads is correct and idiomatic.
- **Well-structured PR description:** Clear root cause, commit-by-commit breakdown, and test plan.

---

## Final Verdict

**Needs Rework** — two issues should be fixed before merge:

| Priority | Issue | Effort |
|----------|-------|--------|
| **Must fix** | #2: `RoundTrip` must clone the request before mutating headers | Small (~5 lines) |
| **Must fix** | #1: Sanitize `filePath` in `Content-Disposition` header | Small (~3 lines) |
| **Should fix** | #3: Investigate whether auth header in URL response is redundant | Investigation |
| **Consider** | #4: Document proxy header trust assumption | Documentation |
| **Consider** | #5: Set `Content-Length` on raw downloads | Small (~1 line) |

mfreeman451:
Addressed the must-fix review items locally, but I cannot push directly to the PR head branch because `Empiria/scion` denied write access for `mfreeman451` even though maintainer edits are enabled.\n\nPrepared fix commit: `dd1a246a`\nFallback branch on my fork: `mfreeman451:review/pr102-local-storage-upload`\n\nChanges in that branch:\n- clone the request before mutating headers in `authRoundTripper.RoundTrip`\n- remove serialized `Authorization` headers from rewritten local upload/download URL responses\n- document the trusted-proxy assumption in `requestBaseURL`\n- sanitize `Content-Disposition` with `mime.FormatMediaType` and `filepath.Base`\n- set `Content-Length` on raw downloads\n- close `r.Body` in the raw upload path\n\nValidation:\n- `go test ./pkg/apiclient -count=1`\n- `go test ./pkg/hub -run ^ -count=1`
- `go test ./pkg/hubclient -count=1`

meatballs:
Thanks @mfreeman451 — I've cherry-picked your review-fix commit (`dd1a246a`) onto the PR head as `5f65744f`. Apologies for the maintainer-edits friction; I've enabled 'Allow edits by maintainers' going forward. All five review items are now addressed on the branch:

- Request cloned before header mutation in `authRoundTripper.RoundTrip`
- `Authorization` header dropped from rewritten local upload/download URL responses
- Trusted-proxy assumption documented in `requestBaseURL`
- `Content-Disposition` sanitized via `mime.FormatMediaType` + `filepath.Base`
- `Content-Length` set on raw downloads; `r.Body` closed in raw upload path

ptone:
merged commits directly after fixing lint

PR #103, opened 2026-04-09
fix: pass dispatch InlineConfig to GetAgent so volumes are applied

### LINKED ISSUE #101: Hub-dispatched agents do not apply harness config or template volumes
## Problem

When starting an agent via the hub, custom volume mounts defined in either the harness config (`settings.yaml` → `harness_configs.<name>.volumes`) or the agent template (`scion-agent.yaml` → `volumes`) are not applied to the container. Only the standard SCION mounts appear:

- `/home/scion` (agent home)
- `/repo-root/.git` (project git)
- `/repo-root/.scion/agents/<name>/workspace` (agent workspace)

Custom volumes like a knowledge vault mount or model cache are silently dropped.

## Reproduction

1. Define volumes in a harness config:
   ```yaml
   harness_configs:
     claude-anvil:
       harness: claude
       image: ghcr.io/empiria/empiria-claude-anvil:latest
       volumes:
         - source: /home/user/vault
           target: /vault
         - source: /home/user/.cache/models
           target: /home/scion/.cache/models
           read_only: true
   ```

2. Start an agent via the hub:
   ```bash
   scion start my-agent -a --image ghcr.io/empiria/empiria-claude-anvil:latest
   ```

3. Inspect the running container:
   ```bash
   podman inspect <container> --format '{{json .Mounts}}'
   ```

4. Observe: only the 3 standard mounts are present. The vault and model cache volumes are missing.

Also tested with `--config <template.yaml>` which passes the template's volume definitions — same result.

## Expected behaviour

Volumes from the harness config and/or template should be applied to the container when the broker creates it, in addition to the standard SCION mounts. If this is intentionally not supported for security reasons (isolation between broker and agent), this should be documented and an alternative mechanism provided (e.g. shared directories, hub-managed volumes).

## Impact

This prevents any agent setup that requires host mounts beyond the project repo — e.g. knowledge vaults, model caches, SSH agent forwarding, or shared tooling. Combined with the template sync issue (#100), custom agent configurations cannot be used with hub-dispatched agents at all.

## Workaround

Use `--no-hub` for local mode, where template volumes are applied correctly. This loses hub UI visibility.

## Environment

- SCION commit a9fca6cb
- Podman 4.9.3 (rootless)
- Linux 6.8.0

meatballs:
## Summary

Fixes #101 — Hub-dispatched agents do not apply harness config or template volumes.

**Root cause:** When `AgentManager.Start()` builds the `startInlineConfig` to pass to `GetAgent()`, it only extracts `HarnessAuth` from `opts`, discarding `opts.InlineConfig` entirely. The InlineConfig carries volumes, env vars, image, and other settings from templates and harness configs that the hub sends via the dispatch request.

The broker correctly receives the InlineConfig from the hub (via `buildStartContext` in `start_context.go`), and `GetAgent` already merges it via `config.MergeScionConfig` — but the connection between the two was missing.

**Fix:** Use `opts.InlineConfig` as the base for `startInlineConfig`, then overlay `HarnessAuth` on top. This is a 7-line change in `pkg/agent/run.go`.

### Before
```go
var startInlineConfig *api.ScionConfig
if opts.HarnessAuth != "" {
    startInlineConfig = &api.ScionConfig{AuthSelectedType: opts.HarnessAuth}
}
```

### After
```go
var startInlineConfig *api.ScionConfig
if opts.InlineConfig != nil {
    startInlineConfig = opts.InlineConfig
}
if opts.HarnessAuth != "" {
    if startInlineConfig == nil {
        startInlineConfig = &api.ScionConfig{}
    }
    startInlineConfig.AuthSelectedType = opts.HarnessAuth
}
```

## Test plan

- [x] `go build ./pkg/agent/...` — compiles cleanly
- [x] `go vet ./pkg/agent/...` — no warnings
- [x] Tested with self-hosted hub + remote broker — custom volumes (vault mount, SSH agent, model cache) now appear in the container
- [ ] Verify local mode (`--no-hub`) still works (InlineConfig is nil, no change in behavior)
- [ ] Verify HarnessAuth override still takes precedence

### DIFF
--- a/pkg/agent/run.go
+++ b/pkg/agent/run.go
@@ -82,12 +82,19 @@ func (m *AgentManager) Start(ctx context.Context, opts api.StartOptions) (*api.A
 		ctx = api.ContextWithGitClone(ctx, opts.GitClone)
 	}
 
-	// Build inline config for GetAgent, ensuring --harness-auth is applied
-	// before harness Provision() runs (which reads auth_selectedType to decide
-	// which env vars to inject).
+	// Build inline config for GetAgent by merging the dispatch InlineConfig
+	// (which carries volumes, env, image, etc. from templates/harness configs)
+	// with any --harness-auth override. Without this merge, custom volumes
+	// from hub-dispatched agents are silently dropped.
 	var startInlineConfig *api.ScionConfig
+	if opts.InlineConfig != nil {
+		startInlineConfig = opts.InlineConfig
+	}
 	if opts.HarnessAuth != "" {
-		startInlineConfig = &api.ScionConfig{AuthSelectedType: opts.HarnessAuth}
+		if startInlineConfig == nil {
+			startInlineConfig = &api.ScionConfig{}
+		}
+		startInlineConfig.AuthSelectedType = opts.HarnessAuth
 	}
 
 	util.Debugf("Start: calling GetAgent name=%s template=%q image=%q harnessConfig=%q grovePath=%q profile=%q",


ptone:
# PR #103 Review: fix: pass dispatch InlineConfig to GetAgent so volumes are applied

**Reviewer:** Scion Code Review Agent
**Date:** 2026-04-11
**Branch:** `fix/dispatch-inline-config-volumes`
**Files changed:** `pkg/agent/run.go` (+11 / -4)

---

## Executive Summary

**Risk Level: Low.** This is a focused 7-line fix that aligns `AgentManager.Start()` with the existing pattern already used in `AgentManager.Create()` (via `ProvisionAgent` at `provision.go:184-190`). The change is structurally sound and addresses a clear data-flow gap where `opts.InlineConfig` was being silently discarded on the `Start` path.

---

## Critical Issues

**None.**

---

## Observations

### 1. Aliasing vs. Copying `opts.InlineConfig` — Acceptable but Worth Noting

```go
if opts.InlineConfig != nil {
    startInlineConfig = opts.InlineConfig  // pointer alias, not copy
}
```

This assigns the pointer directly rather than making a shallow copy. Line 97 then mutates the struct in-place:

```go
startInlineConfig.AuthSelectedType = opts.HarnessAuth
```

This means **`opts.InlineConfig.AuthSelectedType` is mutated as a side-effect** when `HarnessAuth` is non-empty.

**Verdict: Not a bug in practice.** The identical pattern already exists in `ProvisionAgent` (`provision.go:184-189`) where `inlineCfg := opts.InlineConfig` followed by `inlineCfg.AuthSelectedType = opts.HarnessAuth` performs the same aliased mutation. The `opts` struct is not reused after this call in either code path, so the mutation is harmless. However, if `StartOptions` is ever reused (e.g., retry logic), this would become a latent bug.

**Optional defensive fix** (not blocking):
```go
if opts.InlineConfig != nil {
    copied := *opts.InlineConfig
    startInlineConfig = &copied
}
```

### 2. `GetAgent` Receives the Config via Variadic — Correct Usage

`GetAgent` accepts `inlineConfig ...*api.ScionConfig` and indexes `inlineConfig[0]` at `provision.go:1019-1020`, then passes it through to `ProvisionAgent` which calls `MergeScionConfig(finalScionCfg, inlineCfg)`. This merge is additive (override-wins), so passing the full `InlineConfig` (with volumes, env, image, etc.) is exactly the right thing to do — these fields will layer on top of template defaults.

### 3. Comment Quality

The updated block comment (lines 85-88) is clear and explains the *why* — specifically calling out the silent-drop failure mode. This is good practice.

---

## Positive Feedback

- **Consistency:** The fix mirrors the exact pattern already established in `ProvisionAgent` (lines 184-190), making the two entry points (`Start` vs `Create`) consistent. This is the right approach.
- **Minimal blast radius:** Only the config-assembly block was touched; no signatures, no new dependencies, no behavioral changes to `GetAgent` or `MergeScionConfig`.
- **Good PR description:** The before/after comparison and root-cause analysis in the PR body make the intent unambiguous.

---

## Final Verdict

**Approve.** The change is correct, minimal, and consistent with existing patterns. The pointer-aliasing observation is a minor defensive-programming note, not a blocking concern given identical usage elsewhere in the same package.

PR #106, opened 2026-04-09
fix: resolve hub endpoint from --base-url flag

### LINKED ISSUE #104: Hub endpoint ignores --base-url flag, falls back to localhost:8080
## Problem

When the hub is started with `--base-url https://scion.example.com`, the hub endpoint used for broker communication and container env vars still resolves to `http://localhost:8080`. This causes containers to receive:

```
SCION_HUB_ENDPOINT=https://scion.example.com:8080
```

The `:8080` (internal web port) is appended to the public base URL, making the endpoint unreachable from inside the container.

## Root cause

`resolveHubEndpoint()` in `cmd/server_foreground.go` checks `cfg.Hub.Endpoint` and the `SCION_SERVER_BASE_URL` env var, but does not check the `--base-url` command-line flag (`webBaseURL`). When neither config nor env is set, it falls back to `fmt.Sprintf("http://localhost:%d", webPort)`.

The `--base-url` flag is only used later for the web server's `BaseURL` config, not for the hub endpoint resolution.

## Impact

- Containers cannot reach the hub for status reporting or terminal relay
- Hub web terminal fails silently (connection to the wrong endpoint)
- `scion attach` works from the CLI (connects directly) but the hub UI terminal does not

## Reproduction

1. Start hub with: `scion server start --base-url https://scion.example.com --enable-hub --enable-web`
2. Start an agent via the hub
3. Inside the container: `echo $SCION_HUB_ENDPOINT`
4. Observe: `https://scion.example.com:8080` instead of `https://scion.example.com`

## Workaround

Set `SCION_SERVER_BASE_URL=https://scion.example.com` as an environment variable before starting the hub server. This is checked before the localhost fallback.

## Environment

- SCION commit 2d60b5d7
- Hub deployed via Podman on ARM64, accessed via reverse proxy

meatballs:
## Summary

Fixes #104 — Hub endpoint ignores `--base-url` flag, falls back to `localhost:8080`.

`resolveHubEndpoint()` checks `cfg.Hub.Endpoint` and `SCION_SERVER_BASE_URL` env but ignores the `--base-url` CLI flag. When neither config nor env is set, it falls back to `http://localhost:<webPort>`, causing containers to receive the wrong hub endpoint.

**Fix:** Check `webBaseURL` (the `--base-url` flag) first, before the env var and localhost fallback. One line added.

## Test plan

- [ ] Start hub with `--base-url https://example.com` — verify `SCION_HUB_ENDPOINT` in containers matches
- [ ] Start hub without `--base-url` — verify fallback to localhost still works
- [ ] Start hub with `SCION_SERVER_BASE_URL` env — verify env var still works

### DIFF
--- a/cmd/server_foreground.go
+++ b/cmd/server_foreground.go
@@ -621,33 +621,41 @@ func initDevAuth(cfg *config.GlobalConfig, globalDir string) (string, error) {
 
 // resolveHubEndpoint determines the Hub's public endpoint URL.
 func resolveHubEndpoint(cfg *config.GlobalConfig, brokerSettings *config.Settings) string {
-	hubEndpoint := cfg.Hub.Endpoint
-	if hubEndpoint == "" && enableHub {
-		if webBaseURL != "" {
-			hubEndpoint = strings.TrimRight(webBaseURL, "/")
-			if enableDebug {
-				log.Printf("Hub endpoint resolved from --base-url flag: %s", hubEndpoint)
-			}
-		} else if baseURL := os.Getenv("SCION_SERVER_BASE_URL"); baseURL != "" {
-			hubEndpoint = strings.TrimRight(baseURL, "/")
-			if enableDebug {
-				log.Printf("Hub endpoint resolved from SCION_SERVER_BASE_URL: %s", hubEndpoint)
-			}
-		} else {
-			port := cfg.Hub.Port
-			if enableWeb {
-				port = webPort
-			}
-			hubEndpoint = fmt.Sprintf("http://localhost:%d", port)
-			if enableDebug {
-				log.Printf("Auto-computed hub endpoint for combo mode: %s", hubEndpoint)
-			}
-		}
-	} else if hubEndpoint == "" {
-		hubEndpoint = brokerSettings.GetHubEndpoint()
+	if cfg.Hub.Endpoint != "" {
+		return cfg.Hub.Endpoint
+	}
+
+	if !enableHub {
+		hubEndpoint := brokerSettings.GetHubEndpoint()
 		if hubEndpoint != "" && enableDebug {
 			log.Printf("Hub endpoint resolved from grove settings: %s", hubEndpoint)
 		}
+		return hubEndpoint
+	}
+
+	if webBaseURL != "" {
+		hubEndpoint := strings.TrimRight(webBaseURL, "/")
+		if enableDebug {
+			log.Printf("Hub endpoint resolved from --base-url flag: %s", hubEndpoint)
+		}
+		return hubEndpoint
+	}
+
+	if baseURL := os.Getenv("SCION_SERVER_BASE_URL"); baseURL != "" {
+		hubEndpoint := strings.TrimRight(baseURL, "/")
+		if enableDebug {
+			log.Printf("Hub endpoint resolved from SCION_SERVER_BASE_URL: %s", hubEndpoint)
+		}
+		return hubEndpoint
+	}
+
+	port := cfg.Hub.Port
+	if enableWeb {
+		port = webPort
+	}
+	hubEndpoint := fmt.Sprintf("http://localhost:%d", port)
+	if enableDebug {
+		log.Printf("Auto-computed hub endpoint for combo mode: %s", hubEndpoint)
 	}
 	return hubEndpoint
 }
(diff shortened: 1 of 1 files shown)


cmd/server_foreground.go
```diff
@@ -623,7 +623,12 @@ func initDevAuth(cfg *config.GlobalConfig, globalDir string) (string, error) {
 func resolveHubEndpoint(cfg *config.GlobalConfig, brokerSettings *config.Settings) string {
 	hubEndpoint := cfg.Hub.Endpoint
 	if hubEndpoint == "" && enableHub {
-		if baseURL := os.Getenv("SCION_SERVER_BASE_URL"); baseURL != "" {
+		if webBaseURL != "" {
+			hubEndpoint = strings.TrimRight(webBaseURL, "/")
+			if enableDebug {
+				log.Printf("Hub endpoint resolved from --base-url flag: %s", hubEndpoint)
+			}
+		} else if baseURL := os.Getenv("SCION_SERVER_BASE_URL"); baseURL != "" {
```
mfreeman451: should use guard clauses here and eliminate inline if/else

ptone:
# PR #106 Review: fix: resolve hub endpoint from --base-url flag

**PR:** https://github.com/GoogleCloudPlatform/scion/pull/106
**Files changed:** `cmd/server_foreground.go` (+32 / -19)
**Reviewed:** 2026-04-11

---

## Executive Summary

This is a low-risk, well-scoped bug fix that refactors `resolveHubEndpoint()` into an early-return cascade, adding `webBaseURL` (the `--base-url` CLI flag) as a resolution source. The refactor is semantically faithful to the original logic for all existing code paths while correctly inserting the missing flag check.

---

## Critical Issues

**None.**

---

## Observations

### 1. Subtle behavioral change in the `!enableHub` path (low risk)

**Old code:** When `cfg.Hub.Endpoint == ""` and `enableHub == false`, the old code fell into the `else if` branch and called `brokerSettings.GetHubEndpoint()`.

**New code:** When `cfg.Hub.Endpoint == ""` and `!enableHub`, the function returns `brokerSettings.GetHubEndpoint()` immediately — *same behavior*.

However, in the old code, when `enableHub == true` **and** `cfg.Hub.Endpoint != ""`, the function returned `cfg.Hub.Endpoint` directly (the first assignment `hubEndpoint := cfg.Hub.Endpoint` was non-empty, so neither `if` nor `else if` branch executed). The new code preserves this: `cfg.Hub.Endpoint != ""` returns early on line 624.

**Verdict:** Behavioral equivalence confirmed for all pre-existing paths. No issue.

### 2. Priority ordering is correct and well-documented

The new resolution order is:
1. `cfg.Hub.Endpoint` (explicit config)
2. `brokerSettings.GetHubEndpoint()` (when hub is not enabled — grove-local mode)
3. `webBaseURL` / `--base-url` flag **(new)**
4. `SCION_SERVER_BASE_URL` env var
5. `http://localhost:<port>` fallback

The `--base-url` flag taking precedence over the env var is the correct design: CLI flags should override environment variables in standard 12-factor precedence.

### 3. Consider trimming trailing slash on `cfg.Hub.Endpoint` (minor)

The `webBaseURL` and `SCION_SERVER_BASE_URL` paths both call `strings.TrimRight(…, "/")`, but the `cfg.Hub.Endpoint` early return on line 624-625 does not. If a user sets `hub.endpoint: "https://example.com/"` in config, it will propagate with the trailing slash, which could cause double-slash issues downstream (e.g., `https://example.com//api/v1/...`).

This is pre-existing behavior and not introduced by this PR, so it is non-blocking, but worth noting as a follow-up.

### 4. No unit test coverage for the new `webBaseURL` path

The test plan mentions manual verification, but there is no automated test for the new branch. Since `webBaseURL` is a package-level `var` (set via cobra flags), it could be set in a test and the function invoked directly. This would prevent regressions.

**Suggested follow-up** (non-blocking):
```go
func TestResolveHubEndpoint_BaseURLFlag(t *testing.T) {
    oldWebBaseURL := webBaseURL
    oldEnableHub := enableHub
    defer func() {
        webBaseURL = oldWebBaseURL
        enableHub = oldEnableHub
    }()

    enableHub = true
    webBaseURL = "https://example.com/"

    cfg := &config.GlobalConfig{}
    result := resolveHubEndpoint(cfg, &config.Settings{})
    if result != "https://example.com" {
        t.Errorf("expected trimmed base URL, got %q", result)
    }
}
```

---

## Positive Feedback

- **Excellent refactor.** The early-return style is significantly more readable than the nested `if/else if` it replaces. Each resolution step is self-contained and easy to reason about.
- **Correct precedence.** CLI flag > env var > computed default follows Go CLI conventions.
- **Minimal blast radius.** Single function, single file, no interface changes. The fix is surgical.
- **Good PR description.** Clear problem statement, root cause, and manual test plan.

---

## Final Verdict

**Approve.** The change is correct, minimal, and improves readability. The lack of a unit test for the new path is a minor gap but not blocking given the straightforward nature of the fix.

PR #107, opened 2026-04-09
fix: use podman unshare to clean up rootless home directories

### LINKED ISSUE #105: Agent deletion fails to clean up home directory in podman rootless mode
## Problem

When an agent is deleted and recreated in podman rootless mode, root-owned files are left in `~/.scion/grove-configs/*/home/`. The user must manually `sudo rm -rf` the directory before each agent start.

## Root cause

In podman rootless mode, the host user's UID is mapped to container UID 0. Files created as root inside the container (e.g. by sciontool, harness provisioning) are owned by a mapped subuid on the host (e.g. UID 100999). The normal host user cannot remove these files.

`DeleteAgentFiles()` in `pkg/agent/provision.go` calls `util.RemoveAllSafe()` on the external agent home directory, but this fails silently because the host user lacks permission to remove the mapped-UID files. The failure is logged at debug level but not surfaced to the user.

The K8s runtime (`pkg/runtime/k8s_runtime.go:302-308`) already handles this by running `chown -R scion:scion` inside the container before cleanup. The Podman runtime does not have an equivalent.

## Impact

Every agent delete/recreate cycle requires manual `sudo rm -rf` of the home directory. Without this, agent creation fails with "permission denied" when copying harness-config files.

## Reproduction

1. Create and start an agent with podman rootless
2. Delete the agent: `scion delete <agent>`
3. Start a new agent with the same name
4. Observe: fails with permission denied on the home directory
5. Workaround: `sudo rm -rf ~/.scion/grove-configs/<grove>/home`

## Suggested fix

When `RemoveAllSafe()` fails on the external home directory, fall back to `podman unshare rm -rf <path>`. This enters the user namespace where the mapped UIDs are accessible, allowing cleanup without requiring sudo.

## Environment

- Podman 4.9.3 (rootless)
- Linux 6.8.0

meatballs:
## Summary

Fixes #105 — Agent deletion fails to clean up home directory in podman rootless mode.

In podman rootless mode, files created as root inside the container are owned by a mapped subuid on the host. `DeleteAgentFiles()` calls `RemoveAllSafe()` on the external home directory, but this fails silently because the host user can't remove mapped-UID files.

**Fix:** When `RemoveAllSafe()` fails, fall back to `podman unshare rm -rf <path>`. This enters the user namespace where the mapped UIDs are accessible, allowing cleanup without sudo.

This is the Podman equivalent of the `chown -R` pattern already used in the K8s runtime (`pkg/runtime/k8s_runtime.go:302-308`).

## Test plan

- [ ] Delete agent in podman rootless mode — verify home directory is cleaned up
- [ ] Delete agent on Docker or podman rootful — verify standard removal still works (podman unshare not called)
- [ ] Verify `podman unshare` failure is non-fatal (graceful degradation if podman not available)

### DIFF
--- a/pkg/agent/provision.go
+++ b/pkg/agent/provision.go
@@ -127,9 +127,11 @@ func DeleteAgentFiles(agentName string, grovePath string, removeBranch bool) (bo
 			util.Debugf("delete: removing external agent home: %s", externalAgentDir)
 			if err := util.RemoveAllSafe(externalAgentDir); err != nil {
 				util.Debugf("delete: standard removal failed, trying podman unshare: %v", err)
-				if unshareErr := exec.Command("podman", "unshare", "rm", "-rf", externalAgentDir).Run(); unshareErr != nil {
+				unshareCtx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+				if unshareErr := exec.CommandContext(unshareCtx, "podman", "unshare", "rm", "-rf", externalAgentDir).Run(); unshareErr != nil {
 					util.Debugf("delete: podman unshare removal also failed: %v", unshareErr)
 				}
+				cancel()
 			}
 		}
 	}
--- a/pkg/agent/provision.go
+++ b/pkg/agent/provision.go
@@ -19,6 +19,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"strings"
 	"time"
@@ -117,11 +118,18 @@ func DeleteAgentFiles(agentName string, grovePath string, removeBranch bool) (bo
 	}
 
 	// Phase 3: remove external agent home (git grove split storage).
+	// In podman rootless mode, files created as root inside the container are
+	// owned by a mapped subuid on the host, making them inaccessible to the
+	// normal user. If standard removal fails, try `podman unshare rm -rf`
+	// which enters the user namespace where the mapped UIDs are accessible.
 	if externalAgentDir != "" {
 		if _, err := os.Stat(externalAgentDir); err == nil {
 			util.Debugf("delete: removing external agent home: %s", externalAgentDir)
 			if err := util.RemoveAllSafe(externalAgentDir); err != nil {
-				util.Debugf("delete: external home removal failed: %v", err)
+				util.Debugf("delete: standard removal failed, trying podman unshare: %v", err)
+				if unshareErr := exec.Command("podman", "unshare", "rm", "-rf", externalAgentDir).Run(); unshareErr != nil {
+					util.Debugf("delete: podman unshare removal also failed: %v", unshareErr)
+				}
 			}
 		}
 	}


ptone:
# PR #107 Review: fix: use podman unshare to clean up rootless home directories

## Executive Summary

This PR makes two changes: (1) adds a `podman unshare rm -rf` fallback for cleaning up external agent home directories that fail standard removal due to podman rootless UID mapping, and (2) simplifies `GetAgent` by removing worktree recreation logic and relocating the stale-directory cleanup. The primary fix is low-risk and well-scoped; the `GetAgent` refactor removes defensive behavior that may have edge-case implications.

---

## Critical Issues

### 1. Command Injection Risk via `externalAgentDir` — **Low severity (mitigated)**

```go
exec.Command("podman", "unshare", "rm", "-rf", externalAgentDir).Run()
```

`exec.Command` passes arguments as a list (not through a shell), so shell injection via metacharacters in `externalAgentDir` is not possible. Additionally, `externalAgentDir` is constructed from `GetGitGroveExternalAgentsDir(projectDir) + agentName`, where the external dir path comes from on-disk grove config and `agentName` is a function parameter (typically slugified). **No action required**, but worth noting for future maintainers that this path should never be user-controlled raw input passed to a shell.

### 2. Stale Directory Cleanup Lost Worktree Pruning — **Medium severity**

The old code in `GetAgent` pruned worktrees after removing a stale directory:

```go
// OLD
os.RemoveAll(agentDir)
if root, rootErr := util.RepoRootDir(filepath.Dir(agentWorkspace)); rootErr == nil {
    _ = util.PruneWorktreesIn(root)
}
```

The new code removes the directory but **does not prune worktrees**:

```go
// NEW
os.RemoveAll(agentDir)
// (no prune)
```

When a stale agent directory contained a worktree pointer (`.git` file), removing the directory without pruning leaves an orphaned worktree record in `.git/worktrees/`. The subsequent `ProvisionAgent` call does prune worktrees before creating a new one (line 376-380), so this is **likely safe in practice**, but the pruning now happens later in the flow rather than immediately after the stale directory removal. If `ProvisionAgent` fails before reaching the prune step, the orphaned record persists.

**Suggested fix:** Add the prune call back after `os.RemoveAll(agentDir)` in the stale directory block, or document why it's intentionally deferred.

---

## Observations

### 3. Workspace Recreation Logic Removed — Behavioral Change

The old `GetAgent` had ~30 lines of logic to recreate missing worktrees for existing, fully-provisioned agents. The new code simply sets `agentWorkspace = ""` if the workspace doesn't exist. This is a significant behavioral change — previously, a missing workspace would be automatically recreated; now, the caller receives an empty workspace path.

This simplification may be intentional (perhaps the caller handles this case), but it means that if a worktree is accidentally deleted while the agent config still exists, the agent will no longer self-heal on resume. Verify that downstream consumers of `GetAgent` handle `agentWorkspace == ""` gracefully for git-backed agents.

### 4. `podman unshare` Failure is Silent

The fallback logs via `util.Debugf` but does not return an error. This matches the PR's stated design ("graceful degradation"), which is appropriate since the original code also silently logged the failure. The behavior is consistent.

### 5. No Timeout on `exec.Command`

```go
exec.Command("podman", "unshare", "rm", "-rf", externalAgentDir).Run()
```

`Run()` blocks indefinitely. If `podman unshare` hangs (e.g., waiting for a user namespace lock), `DeleteAgentFiles` will never return. Consider using `CommandContext` with a timeout:

```go
ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
defer cancel()
if unshareErr := exec.CommandContext(ctx, "podman", "unshare", "rm", "-rf", externalAgentDir).Run(); unshareErr != nil {
    util.Debugf("delete: podman unshare removal also failed: %v", unshareErr)
}
```

This would require adding `context` to the imports (it's already imported elsewhere in the file via function parameters but `DeleteAgentFiles` doesn't currently take a context).

---

## Positive Feedback

- The comment block explaining the podman rootless UID mapping issue is clear and provides good context for future maintainers.
- The fallback approach (try standard removal first, then podman unshare) is the correct pattern — it avoids unnecessary subprocess invocation on Docker/rootful setups.
- Using `exec.Command` with argument list (not `exec.Command("sh", "-c", ...)`) correctly prevents shell injection.
- The PR description is thorough and references the analogous K8s pattern.

---

## Final Verdict

**Approve with minor suggestions.** The core fix (podman unshare fallback) is correct and low-risk. The two suggestions worth considering before merge:

1. **Add a timeout** to the `exec.Command` call to prevent indefinite hangs (observation #5).
2. **Verify** that the removed worktree recreation logic in `GetAgent` is truly unnecessary, or add the worktree prune back after stale directory removal (issue #2).

Neither is a blocking issue — the PR achieves its stated goal and degrades gracefully on failure.

ptone:
@meatballs I'm happy to patch up minor found issues like this, or let you fix - your choice, just let me know

meatballs:
Thanks for the review @ptone — pushed `ac056d25` which adds a 30s `context.WithTimeout` to the `podman unshare` exec (observation #5). Prevents `DeleteAgentFiles` from blocking indefinitely if the user namespace lock hangs.

Re issues #2 and #3 (worktree prune in `GetAgent`, workspace recreation logic): I think those may have been picked up from combined-branch context rather than this PR. The PR diff only touches `DeleteAgentFiles` in `provision.go` — `GetAgent` is untouched here. Happy to file a follow-up if you'd like the worktree-prune cleanup addressed separately.

PR #110, opened 2026-04-09
Skip metadata interception in non-root containers

mfreeman451:
## Summary
- skip metadata IP interception setup when sciontool is running as a non-root user
- keep metadata serving behavior unchanged while avoiding misleading interception failures in restricted containers
- add focused coverage for the interception decision helper

## Problem
In non-root containers, sciontool cannot install the iptables-based metadata interception rules. The metadata env vars are still the primary mechanism there, so attempting interception just produces noisy, misleading warnings without helping agent behavior.

## Validation
- `go test ./pkg/sciontool/metadata`

### DIFF
--- a/pkg/sciontool/metadata/server.go
+++ b/pkg/sciontool/metadata/server.go
@@ -54,6 +54,11 @@ type Config struct {
 	TokenFunc func() string
 }
 
+const (
+	modeBlock  = "block"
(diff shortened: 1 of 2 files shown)


mfreeman451 (commented):
lgtm

mfreeman451 (commented):
lgtm

PR #111, opened 2026-04-09
Mark failed local agent launches as error

mfreeman451:
## Summary
- mark provisioned local agents as `error` when the runtime launch fails after `agent-info.json` has already been written
- preserve the provisioned workspace/home for diagnosis instead of leaving a phantom `created` agent in list/status output
- add focused regression coverage for the launch-failure state transition

## Problem
The local `scion start` path provisions agent state before attempting the runtime launch. If `Runtime.Run` then fails, the provisioned `agent-info.json` stays in `created` state forever even though no container ever started. In practice that leaves stale phantom agents behind in list output after launch failures.

## Validation
- `go test ./pkg/agent -run 'TestStart_(ErrorPropagation_Tmux|ErrorPropagation_Tmux_Missing|RunFailureMarksAgentInfoError)$'`

### DIFF
--- a/pkg/agent/error_prop_test.go
+++ b/pkg/agent/error_prop_test.go
@@ -17,6 +17,7 @@ package agent
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


mfreeman451 (commented):
lgtm

mfreeman451 (commented):
lgtm

mfreeman451:
Follow-up cleanup is in.

Changes:
- stopped silently dropping local config/status update failures and now log them at debug level
- stopped silently dropping the `prompt.md` and `scion-agent.json` writes
- replaced the inline stringly tmux classification with explicit runtime-error classification
- primary missing-binary detection now uses `errors.Is(err, exec.ErrNotFound)`, with only a narrow fallback for shell-generated `tmux: command not found` text
- updated focused tests accordingly

Current head: `3373b758`

PR #115, opened 2026-04-10
Skip image registry checks for Hub commands

mfreeman451:
## Summary
- skip `image_registry` preflight validation for the `scion hub` command tree
- keep container-launch commands gated on `image_registry` as before
- add a regression test for `scion hub auth login` in a clean global setup

## Testing
- `go test ./cmd -run 'Test(HubAuthLoginDoesNotRequireImageRegistry|FormatFlagCheck|DevAuthWarning|NonInteractiveImpliesAutoConfirm)$'`\n

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -145,7 +145,10 @@ return an error instead of blocking.`,
 		case "config", "doctor":
 			requiresRegistry = false
 		}
-		if parentName == "config" {
+		if commandInSubtree(cmd, "config") {
+			requiresRegistry = false
+		}
+		if commandInSubtree(cmd, "hub") {
 			requiresRegistry = false
 		}
 		if requiresRegistry && config.IsHubContext() {
@@ -196,6 +199,15 @@ func Execute() {
 		}
 		os.Exit(1)
 	}
+}
+
+func commandInSubtree(cmd *cobra.Command, name string) bool {
+	for current := cmd; current != nil; current = current.Parent() {
+		if current.Name() == name {
+			return true
+		}
+	}
+	return false
 }
 
 func init() {
--- a/cmd/root_test.go
+++ b/cmd/root_test.go
@@ -129,6 +129,47 @@ func TestFormatFlagCheck(t *testing.T) {
 	}
 }
 
+func TestHubAuthLoginDoesNotRequireImageRegistry(t *testing.T) {
+	origGlobalMode := globalMode
+	origGrovePath := grovePath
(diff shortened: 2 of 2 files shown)


mfreeman451 (commented):
lgtm

ptone:
# PR #115: Skip image registry checks for Hub commands

## Executive Summary

Low-risk, well-scoped fix that exempts the entire `hub` command subtree from `image_registry` preflight validation, fixing a bug where `scion hub auth login` would fail in a clean global setup. The implementation is correct and introduces a clean helper function with an appropriate regression test.

## Critical Issues

None.

## Observations

### 1. Consolidation opportunity for the `config` subtree check (Minor)

The old code used `parentName == "config"` (single-level check). The PR replaces it with `commandInSubtree(cmd, "config")` which walks the full ancestry — a strict improvement. However, there is now a redundancy: the `switch` on line 144–146 already handles `cmdName == "config"` (the config root itself), while `commandInSubtree` also matches when `current.Name() == "config"` for the config root. This is harmless (it just sets `requiresRegistry = false` twice) but could be noted for future cleanup.

### 2. `commandInSubtree` also matches the root command name

`commandInSubtree` walks all the way to the root (whose `Name()` is typically `"scion"`). If a future developer adds a top-level command named `"hub"`, it would be indistinguishable from the subtree `hub`. This is a non-issue today given the existing command tree structure, but worth noting as a design caveat.

### 3. Test relies on package-level global state (Acceptable)

The test manually saves/restores 8 package-level variables. This pattern already exists in the codebase (`TestDevAuthWarning`, `TestFormatFlagCheck`) so it is consistent. The `defer` restore block is correct.

### 4. Test does not verify the _negative_ case

The test confirms `hub auth login` does **not** require image registry, but does not assert that a command that _should_ require it still fails without one. This is covered by existing tests, so not blocking, but a paired negative assertion would strengthen the regression test.

## Positive Feedback

- **`commandInSubtree` helper** is a clean, reusable utility — better than the previous single-level `parentName` comparison. It correctly handles arbitrary nesting depth.
- The PR description includes an exact `go test` invocation, making verification easy.
- The test sets up a realistic minimal environment (temp `HOME`, global mode, hub endpoint) rather than mocking, which catches integration-level regressions.
- The change is minimal and surgical — only the lines that matter are touched.

## Final Verdict

**Approve.** The change is correct, minimal, well-tested, and introduces no security or performance concerns. Ship it.

PR #124, opened 2026-04-10
fix: derive agent phase from container status in runtime List methods

### LINKED ISSUE #123: Runtime List() hardcodes Phase: created, causing heartbeat to revert running agents
## Summary

The `List()` methods in podman, docker, and apple container runtimes hardcode `Phase: "created"` for all containers. `AgentManager.List()` overrides this from `agent-info.json`, but when the Claude harness overwrites that file with only the activity field (dropping the phase), the heartbeat sends `phase: "created"` to the hub, reverting running agents to `created` in the database.

## Reproduction

1. Start an agent via `scion start` — hub shows `running`
2. Connect to the agent terminal and interact with it (triggering the harness to update `agent-info.json`)
3. The harness writes `{"activity": "thinking"}` to `agent-info.json`, dropping the `phase` field
4. Next heartbeat: `AgentManager.List()` reads the file, finds no phase, falls back to podman's hardcoded `"created"`
5. Hub receives `phase: "created"` and overwrites the `running` phase
6. Hub UI shows agent as `created` — terminal button disabled, agent unreachable

## Root cause

`pkg/runtime/podman.go:272`, `pkg/runtime/docker.go:188`, `pkg/runtime/apple_container.go:214` all set `Phase: "created"` unconditionally. The comments say "Default phase, updated by AgentManager logic" but `AgentManager` depends on `agent-info.json` which can be overwritten by the harness.

## Fix

Derive the phase from the container status string (`Up ...` → `running`, `Exited ...` → `stopped`, else `created`). This matches the hub's own legacy heartbeat derivation logic at `pkg/hub/handlers.go:5280-5293`.

meatballs:
## Summary

- Podman, Docker, and Apple container runtimes hardcode `Phase: "created"` for all containers in `List()`
- When `agent-info.json` loses its phase field (e.g. harness overwrites it with only the activity), the heartbeat sends `created` to the hub, reverting running agents
- Derive phase from container status instead: `Up ...` → `running`, `Exited ...` → `stopped`, else `created`
- Matches the hub's own legacy derivation logic at `pkg/hub/handlers.go:5280-5293`

Fixes #123

## Test plan

- [ ] Start an agent, verify hub shows `running`
- [ ] Interact with agent terminal (triggering harness to overwrite `agent-info.json`)
- [ ] Verify phase stays `running` through subsequent heartbeats
- [ ] Stop agent, verify phase becomes `stopped`

### DIFF
--- a/pkg/runtime/apple_container.go
+++ b/pkg/runtime/apple_container.go
@@ -211,7 +211,7 @@ func (r *AppleContainerRuntime) List(ctx context.Context, labelFilter map[string
 			Labels:          c.Configuration.Labels,
 			Annotations:     c.Configuration.Labels,
 			ContainerStatus: c.Status,
-			Phase:           "created", // Default phase, updated by AgentManager logic
+			Phase:           phaseFromContainerStatus(c.Status),
 			Image:           c.Configuration.Image.Reference,
 			Runtime:         r.Name(),
 		})
--- a/pkg/runtime/common.go
+++ b/pkg/runtime/common.go
@@ -739,3 +739,20 @@ func writeSecretMap(homeDir string, containerHome string, secrets []api.Resolved
 
 	return os.WriteFile(filepath.Join(secretsDir, "secret-map.json"), data, 0600)
 }
+
+// phaseFromContainerStatus derives an agent phase from a container status string.
+// Container runtimes (podman, docker, apple) report status strings like
+// "Up 5 minutes", "Exited (0) 3 hours ago", or "Created". This function
+// maps those to lifecycle phases so heartbeats report accurate state
+// regardless of whether agent-info.json has been updated.
+func phaseFromContainerStatus(status string) string {
+	lower := strings.ToLower(status)
+	switch {
+	case strings.HasPrefix(lower, "up") || lower == "running":
+		return "running"
+	case strings.HasPrefix(lower, "exited") || lower == "stopped":
+		return "stopped"
+	default:
+		return "created"
+	}
+}
--- a/pkg/runtime/docker.go
+++ b/pkg/runtime/docker.go
@@ -185,7 +185,7 @@ func (r *DockerRuntime) List(ctx context.Context, labelFilter map[string]string)
 				ContainerID:     d.ID,
 				Name:            agentName,
 				ContainerStatus: d.Status,
-				Phase:           "created", // Default phase, updated by AgentManager logic
+				Phase:           phaseFromContainerStatus(d.Status),
 				Image:           d.Image,
 				Labels:          labels,
 				Annotations:     labels,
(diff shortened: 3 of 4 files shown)


ptone:
# PR #124 Review: fix: derive agent phase from container status in runtime List methods

## Executive Summary

**Low risk.** This is a small, well-scoped bug fix that replaces hardcoded `"created"` phase values with a helper function that derives phase from container status strings. The logic exactly mirrors the hub's existing legacy derivation at `pkg/hub/handlers.go:5280-5293`, reducing the chance of state regression during heartbeats.

---

## Critical Issues

None.

---

## Observations

### 1. Slight behavioral divergence from hub logic (Minor)

**Location:** `pkg/runtime/common.go:748-758`

The hub's legacy path (handlers.go:5286-5292) treats `containerStatusLower == "created"` specially — it maps to `PhaseProvisioning` (not `"created"`) and includes a guard to avoid downgrading a running agent. The new `phaseFromContainerStatus` function maps the `"created"` status string to the `"created"` phase via the default branch.

This is likely intentional and correct for the `List()` context: a container that genuinely reports `"created"` status hasn't started yet, so `"created"` phase is appropriate. The hub-side guard against downgrading is a server-side concern. However, it's worth noting the difference explicitly, especially if the hub ever starts trusting the client-reported phase without its own derivation.

**Severity:** Informational — no action needed.

### 2. No unit tests for `phaseFromContainerStatus`

The function is pure and deterministic — an ideal candidate for a small table-driven test covering edge cases like:
- `"Up 5 minutes"` → `"running"`
- `"running"` → `"running"`
- `"Exited (0) 3 hours ago"` → `"stopped"`
- `"stopped"` → `"stopped"`
- `"Created"` → `"created"`
- `""` (empty string) → `"created"`
- `"Paused"` → `"created"` (Docker supports paused containers)
- `"Restarting"` → `"created"` (Docker restarting state)

**Severity:** Low — the function is simple and the logic is proven by the hub's existing equivalent, but a test would add confidence for future modifications.

### 3. "Paused" and "Restarting" container states fall to default

Docker containers can be in `Paused` or `Restarting` states. Both will map to `"created"` via the default branch. For `Restarting`, this seems reasonable. For `Paused`, the container is technically still running (just suspended) — mapping to `"created"` could be surprising, though it's unlikely to occur in practice for Scion agents.

**Severity:** Informational.

---

## Positive Feedback

- **Consistency with existing patterns:** The function at line 249 of `apple_container.go` already uses the exact same `strings.HasPrefix(status, "up") && status != "running"` pattern for status checking, confirming the status string formats are well-established.
- **Clean separation:** Extracting the logic into a shared `phaseFromContainerStatus` function in `common.go` avoids duplication across three runtime implementations and centralizes the mapping for future updates.
- **Good documentation:** The function comment clearly explains the purpose and the status string formats it handles.
- **Minimal diff:** The change is surgical — 17 lines added, 3 lines modified, no unnecessary refactoring.

---

## Final Verdict

**Approve.** The fix is correct, minimal, and directly addresses the bug described in #123. The logic mirrors the hub's proven derivation. The only suggestion is adding a table-driven unit test for `phaseFromContainerStatus` to guard against regressions, but this is not blocking.

PR #126, opened 2026-04-10
fix: pre-trust workspace and remove @default model suffix in Claude harness

### LINKED ISSUE #125: Claude harness: trust dialog not pre-accepted, model has invalid @default suffix
## Summary

Two issues in the Claude Code harness prevent unattended agent operation:

### 1. Workspace trust dialog requires manual authorization

`provisionClaudeJSON()` in `pkg/harness/claude_code.go:206` sets `hasTrustDialogAccepted: false` when creating project settings for the container workspace path. The embedded `.claude.json` template has `true`, but since the container workspace path never matches the template's placeholder path, a fresh entry is created with `false`.

This forces every agent to require manual terminal access and authorization before Claude Code will operate, defeating unattended agent execution.

### 2. Model IDs have invalid @default suffix

`pkg/harness/claude/embeds/settings.json` sets:
```json
"ANTHROPIC_MODEL": "claude-opus-4-6@default"
"ANTHROPIC_SMALL_FAST_MODEL": "claude-haiku-4-5@20251001"
```

Claude Code does not recognise the `@provider` suffix format and reports the model as invalid, requiring manual `/model` selection on every session.

## Fix

1. Set `hasTrustDialogAccepted: true` in the fallback project settings
2. Remove `@default` and `@20251001` suffixes from model IDs

meatballs:
## Summary

- Derive container workspace path from `.scion/agents/` directory structure instead of `util.RepoRoot()`. When the broker runs as a systemd service, its CWD is not a git repo, so `RepoRoot()` fails and the trust entry is provisioned for `/workspace` while the actual WORKDIR is `/repo-root/.scion/agents/<name>/workspace`. The mismatch causes Claude to prompt for directory authorization on every session.
- Set `hasTrustDialogAccepted: true` when provisioning new workspace project entries in `.claude.json`.
- Remove `@default` / `@20251001` suffixes from `ANTHROPIC_MODEL` and `ANTHROPIC_SMALL_FAST_MODEL` in embedded `settings.json`. Claude Code does not recognise the `@provider` suffix format and reports the model as invalid.

Fixes #125

## Test plan

- [ ] Start agent via hub with broker running as systemd service
- [ ] Verify no trust dialog on first terminal access
- [ ] Verify Claude uses correct model without manual `/model` selection
- [ ] Verify agent can be used, terminal closed, and re-opened without issues

### DIFF
--- a/pkg/harness/claude/embeds/settings.json
+++ b/pkg/harness/claude/embeds/settings.json
@@ -1,7 +1,7 @@
 {
     "env": {
-    "ANTHROPIC_MODEL":"claude-opus-4-6@default",
-    "ANTHROPIC_SMALL_FAST_MODEL": "claude-haiku-4-5@20251001"
+    "ANTHROPIC_MODEL":"claude-opus-4-6",
+    "ANTHROPIC_SMALL_FAST_MODEL": "claude-haiku-4-5"
   },
   "autoUpdater": {
     "disabled": true
--- a/pkg/harness/claude_code.go
+++ b/pkg/harness/claude_code.go
@@ -173,9 +173,15 @@ func (c *ClaudeCode) provisionClaudeJSON(agentHome, agentWorkspace string) error
 		return err
 	}
 
-	repoRoot, err := util.RepoRoot()
 	containerWorkspace := "/workspace"
-	if err == nil {
+	// Derive the container workspace path from the agent directory structure.
+	// The host path is like .../grove/.scion/agents/<name>/workspace and maps
+	// to /repo-root/.scion/agents/<name>/workspace inside the container.
+	// Use the .scion/agents/ segment as a reliable anchor rather than
+	// util.RepoRoot() which depends on the broker's working directory.
+	if idx := strings.Index(agentWorkspace, "/.scion/agents/"); idx >= 0 {
+		containerWorkspace = "/repo-root" + agentWorkspace[idx:]
+	} else if repoRoot, err := util.RepoRoot(); err == nil {
 		relWorkspace, err := filepath.Rel(repoRoot, agentWorkspace)
 		if err == nil && !strings.HasPrefix(relWorkspace, "..") && relWorkspace != "." {
 			containerWorkspace = filepath.Join("/repo-root", relWorkspace)
@@ -202,7 +208,7 @@ func (c *ClaudeCode) provisionClaudeJSON(agentHome, agentWorkspace string) error
 			"mcpServers":                              map[string]interface{}{},
 			"enabledMcpjsonServers":                   []interface{}{},
 			"disabledMcpjsonServers":                  []interface{}{},
-			"hasTrustDialogAccepted":                  false,
+			"hasTrustDialogAccepted":                  true,
 			"projectOnboardingSeenCount":              1,
 			"hasClaudeMdExternalIncludesApproved":     false,
 			"hasClaudeMdExternalIncludesWarningShown": false,


ptone:
# PR #126 Review: fix: pre-trust workspace and remove @default model suffix in Claude harness

**Branch:** `fix/claude-harness-trust-and-model`
**Files changed:** 2 (+11 / -5)

---

## Executive Summary

This is a low-risk, targeted bug-fix PR that addresses three environment-dependent issues in the Claude Code harness provisioning. The changes are small, well-scoped, and the logic is straightforward with an appropriate fallback chain.

---

## Change Analysis

### 1. `pkg/harness/claude/embeds/settings.json` — Remove `@default`/`@20251001` model suffixes

**Diff:**
```diff
-"ANTHROPIC_MODEL":"claude-opus-4-6@default",
-"ANTHROPIC_SMALL_FAST_MODEL": "claude-haiku-4-5@20251001"
+"ANTHROPIC_MODEL":"claude-opus-4-6",
+"ANTHROPIC_SMALL_FAST_MODEL": "claude-haiku-4-5"
```

**Assessment:** Correct fix. If Claude Code does not recognize the `@provider` suffix format, stripping it is the right call. No issues here.

### 2. `pkg/harness/claude_code.go` — Derive container workspace from path structure

**Diff (core logic):**
```go
if idx := strings.Index(agentWorkspace, "/.scion/agents/"); idx >= 0 {
    containerWorkspace = "/repo-root" + agentWorkspace[idx:]
} else if repoRoot, err := util.RepoRoot(); err == nil {
    ...
}
```

**Assessment:** The logic correctly falls back: try path-based derivation first, then `util.RepoRoot()`, then default `/workspace`. The `strings.Index` approach is simple and reliable.

### 3. `pkg/harness/claude_code.go` — Set `hasTrustDialogAccepted: true`

**Diff:**
```diff
-"hasTrustDialogAccepted": false,
+"hasTrustDialogAccepted": true,
```

**Assessment:** This only applies when creating a *new* project settings entry (the `if projectSettings == nil` branch at line 204). Existing project settings are preserved as-is. Since the harness is provisioning the workspace path itself, pre-trusting it is appropriate and prevents an interactive prompt in a non-interactive context.

---

## Critical Issues

None.

---

## Observations

1. **Path traversal consideration (low risk):** The `agentWorkspace` value is constructed internally by the broker, not from user input. However, if `agentWorkspace` were ever to contain a crafted path like `/malicious/path/.scion/agents/../../etc/`, the resulting `containerWorkspace` would be `/repo-root/.scion/agents/../../etc/`. Since this value is written into `.claude.json` (a trust config), and the input is broker-controlled, this is informational only. If defense-in-depth is desired, a `filepath.Clean()` call on the result would be a minor improvement.

   **Suggested hardening (optional):**
   ```go
   if idx := strings.Index(agentWorkspace, "/.scion/agents/"); idx >= 0 {
       containerWorkspace = filepath.Clean("/repo-root" + agentWorkspace[idx:])
   }
   ```

2. **Test coverage gap:** The existing test (`TestClaudeCode_Provision_VertexAI`) uses `t.TempDir()` for `agentWorkspace`, which will not contain `/.scion/agents/` in its path. This means the new primary code path is untested — it will always fall through to the `util.RepoRoot()` fallback in tests. Consider adding a test case where `agentWorkspace` is set to something like `filepath.Join(tmpDir, ".scion", "agents", "test-agent", "workspace")` to exercise the new branch.

3. **Minor: inconsistent JSON formatting in `settings.json`:** The `ANTHROPIC_MODEL` key has no space after the colon while `ANTHROPIC_SMALL_FAST_MODEL` does. This is pre-existing and not introduced by this PR — noting only for awareness.

---

## Positive Feedback

- The fallback chain (`path anchor -> RepoRoot -> default`) is well-designed and ensures backward compatibility with non-broker invocations.
- The comment block at lines 177–181 clearly explains the reasoning behind the change, which is valuable for future maintainers.
- Scoping `err` to the `else if` branch via `:=` avoids polluting the outer scope — idiomatic Go.

---

## Final Verdict

**Approve.** The changes are correct, well-motivated, and low-risk. The optional `filepath.Clean()` hardening and a dedicated test case for the new path-derivation branch would be nice follow-ups but are not blocking.

PR #129, opened 2026-04-10
Skip workspace bootstrap for global hub starts

mfreeman451:
## Summary
- avoid scanning the global .scion grove as local workspace input during hosted global starts
- keep global hosted start environment gathering scoped to the intended inputs
- add a regression test for the global grove path

## Testing
- go test ./cmd -run 'TestStartAgentViaHub_(EnvGatherFailureCleansUp|GlobalGroveSkipsWorkspaceBootstrap)$' -count=1

### DIFF
--- a/cmd/common.go
+++ b/cmd/common.go
@@ -692,7 +692,7 @@ func startAgentViaHub(hubCtx *HubContext, agentName, task string, resume bool, i
 
 	// Detect non-git grove for workspace bootstrap
 	var workspaceFiles []transfer.FileInfo
-	if hubCtx.GrovePath != "" {
+	if hubCtx.GrovePath != "" && !hubCtx.IsGlobal {
 		groveDir := filepath.Dir(hubCtx.GrovePath) // parent of .scion
 		if _, statErr := os.Stat(groveDir); statErr == nil && !util.IsGitRepoDir(groveDir) {
 			files, err := transfer.CollectFiles(groveDir, transfer.DefaultExcludePatterns)
--- a/cmd/common_envgather_test.go
+++ b/cmd/common_envgather_test.go
@@ -20,7 +20,9 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"os"
+	"path/filepath"
 	"testing"
+	"time"
 
 	"github.com/GoogleCloudPlatform/scion/pkg/hubclient"
 	"github.com/stretchr/testify/assert"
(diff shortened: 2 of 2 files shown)


mfreeman451 (commented):
lgtm

ptone:
# PR #129 Review: Skip workspace bootstrap for global hub starts

**PR:** https://github.com/GoogleCloudPlatform/scion/pull/129
**Branch:** `fix/global-hosted-workspace-collection` -> `main`
**Files Changed:** `cmd/common.go` (+1/-1), `cmd/common_envgather_test.go` (+75/-0)

---

## Executive Summary

This is a low-risk, well-scoped bug fix that adds a single guard condition (`&& !hubCtx.IsGlobal`) to prevent the global `.scion` grove directory from being scanned and uploaded as workspace files during hosted global agent starts. The change is backed by a focused regression test.

## Critical Issues

None.

## Observations

### 1. Test: `GrovePath` set to `settingsPath` rather than grove directory (Minor)

In the new test (`TestStartAgentViaHub_GlobalGroveSkipsWorkspaceBootstrap`), `GrovePath` is set to `settingsPath` (`<globalDir>/settings.yaml`):

```go
hubCtx := &HubContext{
    ...
    GrovePath: settingsPath, // points to settings.yaml
    IsGlobal:  true,
}
```

In the production code at `common.go:690`, `filepath.Dir(hubCtx.GrovePath)` is called to derive the grove directory. So `GrovePath` is expected to be a file *inside* the grove directory. This works correctly here (`filepath.Dir(settingsPath)` == `globalDir`), but the semantics are slightly misleading -- `GrovePath` apparently refers to a file path, not a directory path. This is consistent with existing usage (e.g., `hubsync.UpdateLastSyncedAt(hubCtx.GrovePath, ...)` also passes a file path), so no action needed, but worth noting for documentation.

### 2. Test: global variables saved/restored via `defer` (Style, Non-blocking)

The test manually saves and restores package-level globals (`outputFormat`, `templateName`, `harnessConfigFlag`, `runtimeBrokerID`). This is consistent with the existing `TestStartAgentViaHub_EnvGatherFailureCleansUp` test in the same file, so no objection. The `t.Cleanup` pattern would be slightly more idiomatic Go, but consistency with the existing test is reasonable.

### 3. Test coverage is well-targeted

The test writes a `large-local-file.txt` with content `"should-not-upload"` into the global directory, then asserts `captured.WorkspaceFiles` is empty. This directly validates the fix -- without the `!hubCtx.IsGlobal` guard, `transfer.CollectFiles` would have picked up this file. Good regression test design.

### 4. The fix is correct and minimal

The condition change:
```go
// Before
if hubCtx.GrovePath != "" {
// After  
if hubCtx.GrovePath != "" && !hubCtx.IsGlobal {
```

This is the right place to guard. The `IsGlobal` field is already well-established in the `HubContext` struct (line 161) and used consistently elsewhere (e.g., `stop.go`, `create.go`, `delete.go`). The fix ensures that when running in global scope (e.g., `~/.scion`), the entire home directory structure isn't inadvertently scanned and uploaded as workspace files -- which would be both a performance issue and a potential security concern (accidental exfiltration of unrelated files).

## Positive Feedback

- **Security improvement:** This fix prevents accidental collection and upload of files from the global grove directory (often `~/.scion`), which could include sensitive files in parent directories depending on the grove layout.
- **Minimal blast radius:** One-line production change with a clear, well-structured test.
- **Test design:** The mock HTTP server covers both the grove lookup and agent creation endpoints, and the assertion on `captured.WorkspaceFiles` directly validates the behavioral change.

## Final Verdict

**Approve.** This is a clean, well-tested bug fix with no concerns. The single guard condition is correct, minimal, and addresses a real issue (global grove scanning). Ship it.

PR #130, opened 2026-04-10
Respect explicit Hub endpoint during sync preflight

mfreeman451:
## Summary
- honor an explicit --hub endpoint during sync preflight instead of falling back to settings
- thread the selected endpoint through hub sync resolution
- add focused regression coverage for endpoint override behavior

## Testing
- go test ./pkg/hubsync -run 'TestEnsureHubReady_(GlobalFallbackWithHubEnabled|EndpointOverrideBeatsSettings)$' -count=1
- go test ./cmd -run 'Test.*Flag.*Hub.*|TestGetHubEndpoint.*' -count=1

### DIFF
--- a/cmd/common.go
+++ b/cmd/common.go
@@ -213,6 +213,7 @@ func CheckHubAvailabilityForAgents(grovePath string, excludedAgents []string, sk
 	opts := hubsync.EnsureHubReadyOptions{
 		AutoConfirm:    autoConfirm,
 		NoHub:          noHub,
+		EndpointOverride: hubEndpoint,
 		SkipSync:       skipSync,
 		TargetAgent:    targetAgent,
 		ExcludedAgents: excludedAgents,
--- a/pkg/hubsync/resolve.go
+++ b/pkg/hubsync/resolve.go
@@ -106,7 +106,10 @@ func resolveHubGroveRef(ref string, opts EnsureHubReadyOptions) (*HubContext, er
 			"Enable with: scion config set hub.enabled true")
 	}
 
-	endpoint := getEndpoint(settings)
+	endpoint := opts.EndpointOverride
+	if endpoint == "" {
+		endpoint = getEndpoint(settings)
+	}
 	if endpoint == "" {
 		return nil, fmt.Errorf("hub is enabled but no endpoint configured\n\nConfigure via: scion config set hub.endpoint <url>")
 	}
--- a/pkg/hubsync/sync.go
+++ b/pkg/hubsync/sync.go
@@ -144,6 +144,9 @@ type EnsureHubReadyOptions struct {
 	AutoConfirm bool
 	// NoHub disables Hub integration for this invocation.
 	NoHub bool
+	// EndpointOverride is the explicit Hub endpoint supplied by the caller.
+	// It takes precedence over grove/global settings for this invocation.
+	EndpointOverride string
 	// SkipSync skips agent synchronization check.
 	SkipSync bool
 	// TargetAgent is the agent being operated on. If set, this agent is excluded
@@ -227,7 +230,10 @@ func EnsureHubReady(grovePath string, opts EnsureHubReadyOptions) (*HubContext, 
 	}
 
 	// Hub is enabled - from here on, any failure is an error (no silent fallback)
-	endpoint := getEndpoint(settings)
+	endpoint := opts.EndpointOverride
+	if endpoint == "" {
+		endpoint = getEndpoint(settings)
+	}
 	// In hub context, settings loading may not pick up the env var (e.g. if the
 	// grove path resolves to a synthetic or tmpfs directory without a settings file
 	// and koanf doesn't populate the pointer struct). Fall back to the env var.
--- a/pkg/hubsync/sync_test.go
+++ b/pkg/hubsync/sync_test.go
@@ -112,6 +112,72 @@ hub:
 	}
 }
 
+func TestEnsureHubReady_EndpointOverrideBeatsSettings(t *testing.T) {
+	groveID := "test-override-grove-id"
+
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Content-Type", "application/json")
+		switch {
+		case r.URL.Path == "/healthz":
+			json.NewEncoder(w).Encode(map[string]string{"status": "ok"})
+		case r.URL.Path == "/api/v1/groves/"+groveID:
+			json.NewEncoder(w).Encode(map[string]interface{}{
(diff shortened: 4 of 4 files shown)


mfreeman451 (commented):
lgtm

ncurrie3:
Hi @mfreeman451, great work on this PR! We were working on a similar fix and noticed a couple of edge cases that you might want to consider adding to this PR:

1. In `pkg/hubsync/resolve.go`, the check `if !settings.IsHubEnabled()` might be too strict as it doesn't account for the flag override or container context.
2. Also in `pkg/hubsync/resolve.go`, adding a fallback to `SCION_HUB_ENDPOINT` environment variable when in a Hub context (similar to what you did in `sync.go`).

Here is the snippet we used in `resolve.go` to address both (adapted to use your `EndpointOverride` field name):

```go
	hubContext := config.IsHubContext()
	if opts.EndpointOverride == "" && !settings.IsHubEnabled() && !hubContext {
		return nil, fmt.Errorf("hub grove references (slugs, names, git URLs) require hub mode to be enabled\n\n" +
			"Enable with: scion config set hub.enabled true")
	}

	endpoint := opts.EndpointOverride
	if endpoint == "" {
		endpoint = getEndpoint(settings)
	}
	// Fallback to environment variables in hub-connected containers
	if endpoint == "" && hubContext {
		endpoint = os.Getenv("SCION_HUB_ENDPOINT")
		if endpoint == "" {
			endpoint = os.Getenv("SCION_HUB_URL")
		}
	}
```

Feel free to incorporate this into your PR!

mfreeman451:
> Hi @mfreeman451, great work on this PR! We were working on a similar fix and noticed a couple of edge cases that you might want to consider adding to this PR:
> 
> 1. In `pkg/hubsync/resolve.go`, the check `if !settings.IsHubEnabled()` might be too strict as it doesn't account for the flag override or container context.
> 2. Also in `pkg/hubsync/resolve.go`, adding a fallback to `SCION_HUB_ENDPOINT` environment variable when in a Hub context (similar to what you did in `sync.go`).
> 
> Here is the snippet we used in `resolve.go` to address both (adapted to use your `EndpointOverride` field name):
> 
> ```go
> 	hubContext := config.IsHubContext()
> 	if opts.EndpointOverride == "" && !settings.IsHubEnabled() && !hubContext {
> 		return nil, fmt.Errorf("hub grove references (slugs, names, git URLs) require hub mode to be enabled\n\n" +
> 			"Enable with: scion config set hub.enabled true")
> 	}
> 
> 	endpoint := opts.EndpointOverride
> 	if endpoint == "" {
> 		endpoint = getEndpoint(settings)
> 	}
> 	// Fallback to environment variables in hub-connected containers
> 	if endpoint == "" && hubContext {
> 		endpoint = os.Getenv("SCION_HUB_ENDPOINT")
> 		if endpoint == "" {
> 			endpoint = os.Getenv("SCION_HUB_URL")
> 		}
> 	}
> ```
> 
> Feel free to incorporate this into your PR!

Thanks Nathan, will look this over, appreciate the feedback.

ptone:
# PR #130 Review: Respect explicit Hub endpoint during sync preflight

**Branch:** `fix/hub-flag-endpoint-override`
**Files Changed:** 4 (+78 / -2)
**Reviewer:** Scion Agent (automated)
**Date:** 2026-04-15

---

## Executive Summary

This is a low-risk, well-scoped bug fix that threads an explicit `--hub` endpoint flag through the sync preflight path (`EnsureHubReady` and `resolveHubGroveRef`), ensuring the CLI flag takes precedence over persisted settings. The change is consistent across both call sites and backed by a focused regression test.

## Changes Overview

| File | Change |
|------|--------|
| `pkg/hubsync/sync.go` | Add `EndpointOverride` field to `EnsureHubReadyOptions`; use it in `EnsureHubReady` before falling back to `getEndpoint(settings)` |
| `pkg/hubsync/resolve.go` | Same override-first logic in `resolveHubGroveRef` |
| `cmd/common.go` | Wire `hubEndpoint` (the `--hub` flag value) into `EndpointOverride` |
| `pkg/hubsync/sync_test.go` | New test `TestEnsureHubReady_EndpointOverrideBeatsSettings` |

## Critical Issues

None.

## Observations

### 1. Precedence chain is correct and consistent

Both `EnsureHubReady` (sync.go:233-236) and `resolveHubGroveRef` (resolve.go:109-112) apply the same pattern:

```go
endpoint := opts.EndpointOverride
if endpoint == "" {
    endpoint = getEndpoint(settings)
}
```

In `EnsureHubReady`, the existing env-var fallback (`SCION_HUB_ENDPOINT` / `SCION_HUB_URL`) at line 240 only fires when `endpoint` is still empty *and* we're in a hub container context. The override correctly short-circuits both `getEndpoint()` and the env-var fallback, which is the right behavior for an explicit flag.

### 2. `resolveHubGroveRef` does not have the env-var fallback

In `resolve.go`, after the override check there is only the `getEndpoint(settings)` fallback with no `SCION_HUB_ENDPOINT` env-var fallback (unlike `EnsureHubReady`). This is pre-existing behavior and out of scope, but worth noting: if `resolveHubGroveRef` is ever called from a hub container context without the override, it would fail where `EnsureHubReady` would succeed. This is not a regression introduced by this PR.

### 3. Test quality

The test is well-constructed:
- Stands up an `httptest.Server` as a mock hub with realistic route handling
- Writes a settings file with a **different** endpoint (`http://localhost:8080`) to prove the override wins
- Clears env vars that could interfere (`SCION_HUB_ENDPOINT`, `SCION_AUTH_TOKEN`)
- Asserts the returned `HubContext.Endpoint` matches the override, not settings

Minor nit: the test uses `os.Chdir` + deferred restore. While functional and matching existing test patterns in this file, `t.Chdir()` (available since Go 1.24) would be slightly cleaner. Not blocking.

### 4. No sanitization concern for `EndpointOverride`

The `EndpointOverride` string flows into `createHubClient(settings, endpoint)` which constructs an HTTP client. The value originates from the `--hub` CLI flag (`cmd/root.go:221`), which is a user-controlled input. There is no URL validation on the override, but this is consistent with how `getEndpoint(settings)` values are handled elsewhere -- the health check at the hub endpoint serves as an implicit validation gate. No new attack surface introduced.

## Positive Feedback

- **Minimal, surgical change.** Only 4 production lines changed across 3 files, with clear symmetry between the two call sites.
- **Good field documentation.** The `EndpointOverride` field comment clearly states its semantics and precedence.
- **Test proves the override wins.** The test deliberately sets a conflicting settings endpoint, which is exactly the scenario this bug fixes.

## Final Verdict

**Approve.** The change is correct, consistent, well-tested, and introduces no security or performance concerns.

PR #132, opened 2026-04-10
clean up k8s starter manifests

mfreeman451:
## What changed

This replaces the environment-specific `hack/k8s` deployment snapshot with a generic Kubernetes starter for Scion.

The new base:
- deploys a single combo pod with the Hub, runtime broker, and web UI
- uses a simple `ClusterIP` service plus `kubectl port-forward` as the default access path
- keeps local SQLite and template storage on a PVC
- enables dev auth by default for evaluation so the base can run without preconfigured OAuth
- loads an optional `scion-auth` secret for stable sessions and OAuth credentials

It also removes the CarverAuto-specific ingress, Gateway API, telemetry, sealed secret, pull secret, template seeding, and network policy assumptions.

## Why

The previous manifests were tied to one production deployment and were not suitable for upstream use. They encoded private registry paths, sealed secrets, company-specific hostnames, OTLP mTLS wiring, cluster-specific Gateway resources, and node scheduling details.

That made `hack/k8s` hard to reuse and hard to document as a supported starting point for other Kubernetes users.

## User impact

Users now get a smaller starter that can be applied with `kubectl apply -k hack/k8s` after setting their image and registry values.

Clusters with custom ingress, TLS, private registries, OAuth, or network policy requirements can layer those concerns on top of the base instead of having to unwind an opinionated deployment snapshot first.

## Validation

- `kubectl kustomize hack/k8s`
- `kubectl kustomize hack/k8s | kubectl apply --dry-run=client -f -`

### DIFF
--- a/hack/k8s/.gitignore
+++ b/hack/k8s/.gitignore
@@ -1,0 +1,1 @@
+auth-secret.yaml
--- a/hack/k8s/README.md
+++ b/hack/k8s/README.md
@@ -1,0 +1,84 @@
+This directory contains a generic starter deployment for running Scion in Kubernetes.
+
+The base deploys a single Scion combo pod in the `scion` namespace with:
+
+- Hub API
+- Runtime Broker
+- Web dashboard
+- SQLite and local template storage on a PVC
+- a `ClusterIP` service intended for `kubectl port-forward`
+- development auth enabled by default for evaluation and local testing
+
+The base intentionally does not assume any cluster-specific ingress, TLS, telemetry,
+network policy, private registry, or node scheduling setup.
+
+## Prerequisites
+
+Scion does not currently publish pre-built images. Build and push the images you
+want the cluster to use first.
+
+For a quick start:
+
+```bash
+image-build/scripts/build-images.sh \
+  --registry ghcr.io/<your-org> \
+  --target all \
+  --platform linux/amd64 \
+  --push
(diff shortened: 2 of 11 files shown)


mfreeman451:
Addressed the follow-up cleanup here as well.

Changes:
- aligned the metadata interception setup with the cleaner guard-clause shape already used in `#110`
- extracted the interception setup into `configureMetadataInterception(uid int)`
- kept the earlier manifest fixes in place:
  - `imagePullPolicy: Always` for `:latest`
  - `secrets/get` RBAC
  - broker service port `9800`

Current head: `ecf59882`

mfreeman451:
This branch has been rebuilt and narrowed on top of current `main`. The current diff is now only the starter Kubernetes manifests under `hack/k8s/*`.

The earlier metadata-helper cleanup is already effectively in `main`, so the surviving PR content is just the starter deployment material. The manifest review items remain addressed on the current branch:
- `imagePullPolicy: Always` for `:latest`
- `secrets/get` added to RBAC
- broker service port `9800` exposed

Current branch scope:
- `hack/k8s/.gitignore`
- `hack/k8s/README.md`
- `hack/k8s/*.yaml` starter manifests

Current head: `d520003f`

ptone:
K8s manifests superseded by deploy/helm/

PR #134, opened 2026-04-11
Fallback to GitHub for device flow auth

mfreeman451:
## Summary
- fall back to GitHub when the configured OAuth provider is not available for device flow
- keep the fallback narrowly scoped to the provider-not-configured case
- add focused coverage for the fallback path

## Testing
- go test ./pkg/hub/auth -run "TestDeviceFlowAuth_" -count=1

### DIFF
--- a/cmd/hub_auth.go
+++ b/cmd/hub_auth.go
@@ -136,7 +136,7 @@ func runHubAuthLogin(cmd *cobra.Command, args []string) error {
 	}
 
 	if hubAuthNoBrowser || util.IsHeadlessEnvironment() {
-		provider, err := resolveHubAuthProvider(cmd.Context(), client.Auth(), hubclient.OAuthClientTypeDevice, provider)
+		provider, err := resolveExplicitDeviceFlowProvider(cmd.Context(), client.Auth(), provider)
 		if err != nil {
 			return err
 		}
@@ -155,6 +155,13 @@ func runHubAuthLogin(cmd *cobra.Command, args []string) error {
 	}
 
 	return storeTokenAndPrintResult(hubURL, tokenResp)
+}
+
+func resolveExplicitDeviceFlowProvider(ctx context.Context, authSvc hubclient.AuthService, requestedProvider string) (string, error) {
+	if strings.TrimSpace(requestedProvider) == "" {
+		return "", nil
+	}
+	return resolveHubAuthProvider(ctx, authSvc, hubclient.OAuthClientTypeDevice, requestedProvider)
 }
 
 // runBrowserAuthFlow performs the browser-based OAuth flow.
--- a/cmd/hub_auth.go
+++ b/cmd/hub_auth.go
@@ -141,7 +141,7 @@ func runHubAuthLogin(cmd *cobra.Command, args []string) error {
 			return err
 		}
 		// Device authorization flow for headless environments
-		deviceAuth := auth.NewDeviceFlowAuth(client.Auth(), provider)
+		deviceAuth := newDeviceFlowAuth(client.Auth(), provider)
 		tokenResp, err = deviceAuth.Authenticate(cmd.Context())
 		if err != nil {
 			return fmt.Errorf("device flow authentication failed: %w", err)
@@ -155,6 +155,13 @@ func runHubAuthLogin(cmd *cobra.Command, args []string) error {
 	}
 
 	return storeTokenAndPrintResult(hubURL, tokenResp)
+}
+
+func newDeviceFlowAuth(authSvc hubclient.AuthService, provider string) *auth.DeviceFlowAuth {
+	if strings.TrimSpace(provider) == "" {
+		return auth.NewDeviceFlowAuth(authSvc)
+	}
+	return auth.NewDeviceFlowAuth(authSvc, provider)
 }
 
 func resolveExplicitDeviceFlowProvider(ctx context.Context, authSvc hubclient.AuthService, requestedProvider string) (string, error) {
--- a/cmd/hub_auth_test.go
+++ b/cmd/hub_auth_test.go
@@ -146,3 +146,36 @@ func TestResolveHubAuthProvider_NoProviders(t *testing.T) {
 		t.Fatalf("unexpected error: %v", err)
 	}
 }
+
+func TestResolveExplicitDeviceFlowProvider_ImplicitProviderSkipsResolution(t *testing.T) {
+	t.Parallel()
+
+	authSvc := &mockHubAuthService{
+		providersResp: &hubclient.AuthProvidersResponse{
+			ClientType: string(hubclient.OAuthClientTypeDevice),
+			Providers:  hubclient.OAuthProviderOrder(),
+		},
+	}
+
+	got, err := resolveExplicitDeviceFlowProvider(context.Background(), authSvc, "")
+	if err != nil {
+		t.Fatalf("resolveExplicitDeviceFlowProvider returned error: %v", err)
+	}
+	if got != "" {
+		t.Fatalf("resolveExplicitDeviceFlowProvider = %q, want empty string", got)
+	}
+}
+
+func TestResolveExplicitDeviceFlowProvider_ExplicitProviderUsesValidation(t *testing.T) {
+	t.Parallel()
+
+	authSvc := &mockHubAuthService{}
+
+	got, err := resolveExplicitDeviceFlowProvider(context.Background(), authSvc, "GitHub")
+	if err != nil {
+		t.Fatalf("resolveExplicitDeviceFlowProvider returned error: %v", err)
+	}
+	if got != "github" {
+		t.Fatalf("resolveExplicitDeviceFlowProvider = %q, want github", got)
+	}
+}
--- a/cmd/hub_auth_test.go
+++ b/cmd/hub_auth_test.go
@@ -16,15 +16,21 @@ package cmd
 
 import (
 	"context"
+	"fmt"
 	"strings"
 	"testing"
 
+	"github.com/GoogleCloudPlatform/scion/pkg/apiclient"
 	"github.com/GoogleCloudPlatform/scion/pkg/hubclient"
 )
 
 type mockHubAuthService struct {
-	providersResp *hubclient.AuthProvidersResponse
-	providersErr  error
+	providersResp         *hubclient.AuthProvidersResponse
+	providersErr          error
+	requestDeviceCodeFunc func(ctx context.Context, provider string) (*hubclient.DeviceCodeResponse, error)
+	pollDeviceTokenFunc   func(ctx context.Context, deviceCode, provider string) (*hubclient.DeviceTokenPollResponse, error)
+	requestedProviders    []string
+	polledProviders       []string
 }
 
 type mockHubAuthNotImplementedError struct{}
@@ -58,9 +64,17 @@ func (m *mockHubAuthService) ExchangeCode(ctx context.Context, code, callbackURL
 	return nil, mockHubAuthNotImplementedError{}
 }
 func (m *mockHubAuthService) RequestDeviceCode(ctx context.Context, provider string) (*hubclient.DeviceCodeResponse, error) {
+	m.requestedProviders = append(m.requestedProviders, provider)
+	if m.requestDeviceCodeFunc != nil {
+		return m.requestDeviceCodeFunc(ctx, provider)
+	}
 	return nil, mockHubAuthNotImplementedError{}
 }
 func (m *mockHubAuthService) PollDeviceToken(ctx context.Context, deviceCode, provider string) (*hubclient.DeviceTokenPollResponse, error) {
+	m.polledProviders = append(m.polledProviders, provider)
+	if m.pollDeviceTokenFunc != nil {
+		return m.pollDeviceTokenFunc(ctx, deviceCode, provider)
+	}
 	return nil, mockHubAuthNotImplementedError{}
 }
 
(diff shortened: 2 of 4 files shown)


ptone:
This was partially addressed in #135

See feedback at end of this review:

# Code Review: Issue #69 - OAuth Provider Defaulting to Google

**Issue:** [#69](https://github.com/GoogleCloudPlatform/scion/issues/69) - CLI and server default OAuth provider to Google with no way to override  
**Related PRs:**
- [PR #71](https://github.com/GoogleCloudPlatform/scion/pull/71) - Server-side: `DefaultProviderForClient` fallback in `handlers_auth.go` + `oauth.go`
- [PR #134](https://github.com/GoogleCloudPlatform/scion/pull/134) - Client-side: Device flow provider fallback in `device_flow.go`

**Reviewer:** Scion Agent  
**Date:** 2026-04-11

---

## Executive Summary

These two PRs address issue #69 from complementary sides: PR #71 adds server-side provider auto-detection across all five auth handler endpoints, while PR #134 adds client-side fallback logic specifically for the device flow. Both are low-to-medium risk changes. The main concerns are a fragile error-message-based detection pattern in PR #134 and the potential for redundant fallback logic when both PRs land together.

---

## PR #71 - Server-Side Fix

### Files Changed
- `pkg/hub/handlers_auth.go` (+28/-9)
- `pkg/hub/oauth.go` (+9/-0)

### Analysis

#### Logic & Correctness

**`DefaultProviderForClient` (oauth.go)**
The new method delegates to `ConfiguredProvidersForClient` which iterates providers in stable order (Google first, then GitHub). This preserves backward compatibility while enabling fallback. The terminal fallback to `"google"` when no providers are configured is a reasonable default since the `oauthService` existing but having no configured providers is an unusual state.

**Handler changes (handlers_auth.go)**
All five locations follow a consistent pattern:

```go
if s.oauthService != nil {
    provider = s.oauthService.DefaultProviderForClient(clientType)
} else {
    provider = "google"
}
```

The nil-guard on `oauthService` is good defensive coding.

**One nuance in `handleAuthToken` (line ~327):** The redirect-URI sniffing (`strings.Contains(req.RedirectURI, "github")`) still takes precedence over the new `DefaultProviderForClient` call. This is correct since the redirect URI is a stronger signal, but the `else if` chain means this handler behaves slightly differently from the other four (which have no redirect-URI check). This asymmetry already existed but is worth documenting.

#### Positive Feedback
- Consistent application of the same pattern across all five handler locations.
- Good separation of concerns: the `DefaultProviderForClient` method encapsulates the lookup logic.
- The nil-guard on `oauthService` means the code won't panic in edge cases.

#### Observations
- **No test coverage for the new `DefaultProviderForClient` method.** Consider adding a unit test for `OAuthService.DefaultProviderForClient` verifying (a) Google is returned when both are configured, (b) GitHub is returned when only GitHub is configured, (c) "google" is returned when none are configured.

#### Verdict: **Approve with minor suggestion** (add unit test for `DefaultProviderForClient`).

---

## PR #134 - Client-Side Fix

### Files Changed
- `pkg/hub/auth/device_flow.go` (+47/-14)
- `pkg/hub/auth/device_flow_test.go` (+72/-20)

### Analysis

#### Logic & Correctness

**Variadic constructor signature change:**

```go
// Before
func NewDeviceFlowAuth(client hubclient.AuthService, provider string) *DeviceFlowAuth

// After
func NewDeviceFlowAuth(client hubclient.AuthService, provider ...string) *DeviceFlowAuth
```

This is backward-compatible (existing callers passing a single string still work), but introduces a subtle hazard: a caller passing multiple strings (e.g., `NewDeviceFlowAuth(client, "google", "github")`) would silently use only the first. Consider documenting this or restricting to at most one:

```go
if len(provider) > 1 {
    panic("NewDeviceFlowAuth accepts at most one provider")
}
```

**`providerExplicit` flag:**
The distinction between "caller omitted provider" and "caller passed empty string" is well-thought-out. When `providerExplicit` is true and provider is empty, the original validation error fires. When `providerExplicit` is false (no argument), the fallback path is taken. This is clean.

**`requestDeviceCode` method:**
The fallback loop iterates `hubclient.OAuthProviderOrder()` (Google, then GitHub), trying each until one succeeds. The loop correctly short-circuits on non-provider-configuration errors (line ~138: `if !isProviderNotConfiguredError(err)` returns immediately). This prevents masking unrelated errors (network failures, auth issues, etc.).

#### Critical Issues

**1. Fragile error detection via string matching (High)**

```go
func isProviderNotConfiguredError(err error) bool {
    var apiErr *apiclient.APIError
    return errors.As(err, &apiErr) &&
        apiErr.Code == apiclient.ErrCodeValidationError &&
        strings.Contains(apiErr.Message, "OAuth provider not configured for device flow")
}
```

This couples the client to the exact error message string from the server. If the server message changes (e.g., capitalization, rewording, localization), the fallback silently stops working and returns the error for the first provider tried.

**Suggested Fix:** Add a dedicated error code on the server side (e.g., `ErrCodeProviderNotConfigured`) and check for that instead of parsing the message string. If that's not feasible in this PR, at minimum:
- Add a code comment noting the coupling to the server-side message.
- Consider matching a shorter substring (e.g., `"provider not configured"`) to be more resilient.

**2. Caller already resolves provider (Medium)**

The existing caller in `cmd/hub_auth.go:139-144`:

```go
provider, err := resolveHubAuthProvider(ctx, client.Auth(), hubclient.OAuthClientTypeDevice, provider)
// ...
deviceAuth := auth.NewDeviceFlowAuth(client.Auth(), provider)
```

`resolveHubAuthProvider` already queries the server for available providers. If this function works correctly, the fallback in `requestDeviceCode` would never trigger. This means:
- The fallback is defense-in-depth (good), but the caller would need updating to **not** pass a resolved provider to allow the fallback to activate.
- Or the fallback only helps callers that bypass `resolveHubAuthProvider` (e.g., tests, future callers).

**Action needed:** Verify that the caller in `cmd/hub_auth.go` is updated (or planned to be updated) to call `NewDeviceFlowAuth(client.Auth())` without a provider when the user hasn't explicitly specified one, otherwise this fallback code is unreachable in production.

#### Observations

- The test `TestDeviceFlowAuth_FallsBackToGitHubWhenGoogleUnavailable` is well-structured and directly validates the fallback scenario. Good use of `requestDeviceCodeFunc` / `pollDeviceTokenFunc` hooks in the mock.
- Minor style nits in the test (removing comments like `// 1 second for fast test`, `// No poll responses`, etc.) are acceptable cleanup.
- The `polledProviders` assertion in the test uses manual slice comparison. Could use `reflect.DeepEqual` or `slices.Equal` for clarity, though this is non-blocking.

#### Positive Feedback
- The `requestDeviceCode` helper cleanly separates the retry/fallback logic from the main `Authenticate` flow.
- Error propagation is correct: `lastErr` captures the final provider's error, so the returned error is meaningful.
- Test coverage for the fallback path is thorough.

#### Verdict: **Request changes** - the string-matching error detection is a correctness risk.

---

## Cross-PR Considerations

When both PRs land:
1. **Redundant fallback:** The server (PR #71) will auto-select the right provider when the client sends an empty provider. The client (PR #134) will also try multiple providers. In practice, the client fallback would only trigger if the server doesn't have PR #71's changes. This is fine as defense-in-depth but should be documented.
2. **Provider ordering consistency:** Both PRs use the same provider order (Google first, GitHub second). PR #71 uses `oauthProviderOrder()` (internal), PR #134 uses `hubclient.OAuthProviderOrder()` (public). These should remain in sync.

---

## Summary Table

| Area | PR #71 (Server) | PR #134 (Client) |
|------|-----------------|-------------------|
| Correctness | Good | Good, with caveat on string matching |
| Security | No issues | No issues |
| Test coverage | Missing unit test | Good |
| Architecture | Clean, consistent | Clean but fragile error coupling |
| Risk | Low | Medium |

---

## Final Verdict

- **PR #71:** Approve with minor suggestion to add unit tests for `DefaultProviderForClient`.
- **PR #134:** Request changes on the fragile `isProviderNotConfiguredError` string matching. Recommend introducing a dedicated error code or at minimum documenting the coupling. Also verify the caller in `cmd/hub_auth.go` is updated to not pass a pre-resolved provider, otherwise the fallback is dead code in production.

mfreeman451:
Addressed the client-side production-path concerns here.

Changes:
- `cmd/hub_auth.go` no longer pre-resolves the device-flow provider when the user did not explicitly pass `--provider`, so the fallback path is now reachable in normal headless/device-flow use
- explicit `--provider` still goes through validation via a small `resolveExplicitDeviceFlowProvider(...)` helper
- `NewDeviceFlowAuth(...)` now rejects more than one provider argument
- `isProviderNotConfiguredError(...)` is still temporary message-based matching, but it is now documented as coupled to current server validation wording and matches a shorter/lowercased substring instead of the full sentence
- added focused tests for the explicit-vs-implicit device-flow provider path

Focused validation:
- `go test ./cmd -run 'TestResolve(HubAuthProvider|ExplicitDeviceFlowProvider)_ ' -count=1`\n- `go test ./pkg/hub/auth -run 'TestDeviceFlowAuth_' -count=1`\n\nCurrent head: `1fe70b07`

ptone:
# PR #134 Review: Fallback to GitHub for device flow auth

## Executive Summary

This PR adds a fallback mechanism to the device flow OAuth so that when the configured provider (e.g. Google) is unavailable, authentication falls back to GitHub. **The PR contains a critical integration bug: the fallback path is unreachable from the production call site**, because `NewDeviceFlowAuth` is always called with an explicit (possibly empty) provider string, which prevents the implicit-provider fallback loop from ever executing.

---

## Critical Issues

### 1. Fallback path is unreachable from production code

**Severity: P0 — the stated purpose of this PR does not work end-to-end.**

**Explanation:** When a user runs `hub auth login` in a headless environment _without_ specifying `--provider`:

1. `provider` reads as `""` from the flag (line 133).
2. `resolveExplicitDeviceFlowProvider(ctx, authSvc, "")` returns `("", nil)` — correctly short-circuits.
3. `auth.NewDeviceFlowAuth(client.Auth(), provider)` is called with `""` on line 144.

Inside `NewDeviceFlowAuth(client, "")`:
- `len(provider)` is **1** (variadic receives one element: the empty string).
- `providerExplicit` is set to **`true`**, `selectedProvider` is `""`.

Inside `Authenticate()`:
- `d.providerExplicit && d.provider == ""` evaluates to **`true`** → returns `errDeviceFlowProviderRequired`.

The fallback loop in `requestDeviceCode()` is only entered when `providerExplicit == false`, which requires calling `NewDeviceFlowAuth(client)` with **zero** variadic arguments. The production call site never does this.

The test `TestDeviceFlowAuth_FallsBackToGitHubWhenGoogleUnavailable` calls `NewDeviceFlowAuth(mock)` (zero args), which is not how production calls the constructor.

**Location:** `cmd/hub_auth.go:144`, `pkg/hub/auth/device_flow.go:46-56`

**Suggested Fix (option A — adjust call site):**
```go
// cmd/hub_auth.go, lines 143-144
var deviceAuth *auth.DeviceFlowAuth
if provider != "" {
    deviceAuth = auth.NewDeviceFlowAuth(client.Auth(), provider)
} else {
    deviceAuth = auth.NewDeviceFlowAuth(client.Auth())
}
```

**Suggested Fix (option B — keep constructor simple, move semantics to field):**
Change `NewDeviceFlowAuth` back to a non-variadic signature and use a separate `WithProvider` option or an explicit boolean parameter, avoiding the subtle variadic semantics where `""` and "not provided" are different.

---

### 2. `isProviderNotConfiguredError` is coupled to a server-side message substring

**Severity: Medium — fragile, but acknowledged in the code comment.**

The function matches `strings.Contains(strings.ToLower(apiErr.Message), "provider not configured")`. If the server message changes (e.g., to "provider is not configured" or localized text), the fallback silently stops working and the user gets an opaque error.

**Location:** `pkg/hub/auth/device_flow.go:147-153`

**Suggestion:** This is noted in the code comment as intentional tech debt. At minimum, add a test that asserts the exact error message wording used in production, so breakage is caught early. Consider requesting a dedicated error code from the API team.

---

## Observations

### 3. Variadic constructor is unconventional and error-prone

`NewDeviceFlowAuth(client hubclient.AuthService, provider ...string)` with a `panic` for `len > 1` is an unusual Go pattern. It creates a function that _looks_ like it accepts multiple providers but actually doesn't, and it conflates "zero args" with "empty string arg" in a way that is non-obvious to callers.

A more idiomatic approach:
- Use a functional options pattern: `NewDeviceFlowAuth(client, auth.WithProvider("github"))`.
- Or use two constructors: `NewDeviceFlowAuth(client)` and `NewDeviceFlowAuthWithProvider(client, provider)`.

### 4. `requestDeviceCode` returns `lastErr` which may be `nil`

If `OAuthProviderOrder()` returns an empty slice (which it currently doesn't, but defensively), `requestDeviceCode` returns `nil, "", nil` — a nil error with an empty response. This would cause a nil-pointer dereference when `Authenticate` accesses `codeResp.VerificationURL`.

**Location:** `pkg/hub/auth/device_flow.go:134-147`

**Suggested Fix:**
```go
if lastErr == nil {
    return nil, "", errors.New("no OAuth providers available for device flow")
}
return nil, "", lastErr
```

### 5. Minor: slice comparison in test is fragile

In `TestDeviceFlowAuth_FallsBackToGitHubWhenGoogleUnavailable`, the assertion:
```go
if got, want := mock.requestedProviders, []string{"google", "github"}; len(got) != len(want) || got[0] != want[0] || got[1] != want[1] {
```
will panic with an index-out-of-range if `len(got) < 2` since `||` does not short-circuit array access after the length check in Go (it does short-circuit, but the _assignments_ have already bound `want`). Actually in this case the short-circuit does protect the indexing, but using `reflect.DeepEqual` or `slices.Equal` would be clearer and more maintainable.

---

## Positive Feedback

- **Clean separation of concerns:** Extracting `resolveExplicitDeviceFlowProvider` to distinguish "user explicitly asked for a provider" from "auto-detect" is the right architectural instinct.
- **Good test coverage intent:** The fallback test (`TestDeviceFlowAuth_FallsBackToGitHubWhenGoogleUnavailable`) is well-structured with mock functions and verifies both the request and polling provider sequences.
- **Narrowly scoped fallback:** Only falling back on `provider not configured` (rather than all errors) is the correct defensive approach.

---

## Final Verdict

**Needs rework.** The primary feature (fallback to GitHub when no provider is specified) is not reachable from the production call site due to the variadic constructor semantics. The test passes because it calls the constructor differently than production code does. This must be fixed before merge.

mfreeman451:
Addressed the production-path bug called out in review.\n\nNew head: `1355152d`\n\nChanges:\n- headless device-flow login now calls `NewDeviceFlowAuth(authSvc)` with zero args when the user did not explicitly pass `--provider`\n- explicit `--provider` still goes through `resolveExplicitDeviceFlowProvider(...)` and then uses the one-arg constructor path\n- added a command-level test that exercises the real implicit-provider path and proves fallback requests `google` then `github`\n- existing lower-level `pkg/hub/auth` tests still pass\n\nValidation:\n- `go test ./cmd -run TestResolve(HubAuthProvider|ExplicitDeviceFlowProvider)_|TestNewDeviceFlowAuth_ImplicitProviderUsesFallbackPath -count=1`\n- `go test ./pkg/hub/auth -run TestDeviceFlowAuth_ -count=1`

PR #135, opened 2026-04-11
Fix/oauth provider fallback

ptone:

- [ ] Tests pass


ptone:
@meatballs - I was not able to push the added test to Empiria - so using this to replace #71 - thanks for the contribution

PR #140, opened 2026-04-12
Resolve hosted Kubernetes agent slugs to pod targets

mfreeman451:
## Summary
This fixes hosted Kubernetes runtime operations that still assumed the incoming agent identifier was the pod name.

## Changes
- add `resolvePodTarget(...)` in `pkg/runtime/k8s_runtime.go`
- resolve agent slug to the actual Kubernetes pod name/namespace using `List()` metadata when available
- use that resolver in `GetLogs`, `Attach`, and `Exec`
- keep `namespace/pod` input working as-is
- add focused test coverage for hosted slug -> pod target resolution in `pkg/runtime/k8s_runtime_test.go`

## Why
Hosted agents use a slug like `k8s-reverify-otelenv-233355`, while the actual pod name is prefixed, for example `global--k8s-reverify-otelenv-233355`. Before this change, broker-backed logs and related runtime operations could fail with pod-not-found errors because the runtime tried to use the slug as the pod name.

## Validation
- `go test ./pkg/runtime -run 'TestKubernetesRuntime_(List|List_TerminalPhases|ResolvePodTargetFromHostedSlug)$|TestSelectLogContainer$' -count=1`
- `go test ./pkg/runtimebroker -run 'Test(MethodNotAllowed|AgentLogsAllowsGet)$' -count=1`
- verified live on `scion-int` by rerunning `scion logs <slug> --global --hub http://127.0.0.1:18080` against an existing hosted agent that previously failed with `pods "<slug>" not found`

### DIFF
--- a/pkg/runtime/k8s_runtime.go
+++ b/pkg/runtime/k8s_runtime.go
@@ -26,6 +26,7 @@ import (
 	"os/exec"
 	"os/signal"
 	"path/filepath"
+	"regexp"
 	"strings"
 	"syscall"
 	"time"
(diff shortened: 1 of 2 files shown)


mfreeman451:
Addressed the latest review feedback on this branch.

Changes in the new head (`45dcfd8b`):
- `commandForExec(...)` now uses the pod security context as the cheap primary path and only probes the live pod user when the spec does not already tell us the container runs as non-root
- `resolvePodTarget(...)` now propagates `List(...)` failures instead of silently discarding them, and the redundant post-filtering is gone
- `Attach(...)` now relies on `execTargetUsername(pod)` directly and no longer re-validates the same `scion.username` annotation from the resolved agent metadata
- added focused tests for the non-root exec fast path and the new `resolvePodTarget(...)` error propagation behavior

Validation:
- `go test ./pkg/runtime -run 'TestKubernetesRuntime_(ResolvePodTargetFromHostedSlug|ResolvePodTarget_PropagatesListError|CommandForExec_SkipsProbeForNonRootPod|List|List_TerminalPhases)$|TestSelectLogContainer$'`\n- `go test ./pkg/runtimebroker -run 'Test(MethodNotAllowed|AgentLogsAllowsGet)$'`

ptone:

PR #142, opened 2026-04-12
fix(web): decouple Messages tab from cloudLogging gate

meatballs:
## Summary

The Messages tab on the agent detail page is currently hidden unless the agent has Cloud Logging enabled, even though the primary message source has been the hub store since #54d67ad1.

- The tab was originally added in d2b2ca87 with Cloud Logging as its only data source, so gating on \`cloudLogging\` was correct at the time.
- Commit 54d67ad1 (\"feat(messages): implement Phase 5 web frontend inbox tray\") migrated \`agent-message-viewer\` to use \`GET /api/v1/agents/{id}/messages\` as the primary source, with Cloud Logging as a fallback for pre-migration records. The gate in \`agent-detail.ts\` was not updated at the time.
- The result is that any deployment not opted in to Cloud Logging (per the [observability docs](https://googlecloudplatform.github.io/scion/hub-admin/observability/), Cloud Logging is optional) has no way to see or use the Messages tab, even though the hub-store fetch path works fine and \`tmux send-keys\` delivery via the existing broker path is fully functional.

This PR drops the \`cloudLogging\` gate on the Messages tab specifically. The Logs tab stays gated because \`scion-agent-log-viewer\` only queries Cloud Logging endpoints and has no hub-store equivalent.

## Verification

Tested on a self-hosted hub with no Cloud Logging configured:

- Messages tab appears on the agent detail page as expected
- Sending a message from the compose box reaches the agent via the broker's \`tmux send-keys\` path and the agent responds in its terminal
- Fetch path uses \`/api/v1/agents/{id}/messages\` (hub store) successfully
- Empty state displays correctly when no messages exist

Known limitation not addressed by this PR: the Stream toggle in the viewer hits \`/message-logs/stream\` which returns 501 when \`logQueryService\` is nil. That's a separate migration gap from the same Phase 5 change — worth a follow-up PR to wire the per-agent stream to the existing \`MessageBrokerProxy\` event bus that \`inbox-tray\` already consumes. Happy to open that as a second PR if the approach is agreed.

## Test plan

- [x] Messages tab visible on agent detail page without Cloud Logging
- [x] Send message succeeds and reaches the agent
- [x] Logs tab still gated on \`cloudLogging\` (unchanged)
- [x] No regressions on agents with Cloud Logging enabled (Messages tab still works, Cloud Logging fallback path in viewer still intact)

cc @ptone since both the original tab addition (d2b2ca87) and the Phase 5 migration (54d67ad1) were yours — this is a small cleanup of that migration.

### DIFF
--- a/web/src/components/pages/agent-detail.ts
+++ b/web/src/components/pages/agent-detail.ts
@@ -933,6 +933,7 @@ export class ScionPageAgentDetail extends LitElement {
             agentId=${this.agentId}
             agentName=${this.agent.name || ''}
             ?canSend=${can(this.agent._capabilities, 'message')}
+            ?cloudLogging=${this.agent.cloudLogging || false}
           ></scion-agent-message-viewer>
         </sl-tab-panel>
         <sl-tab-panel name="configuration">${this.renderConfigurationTab()}</sl-tab-panel>
--- a/web/src/components/pages/agent-detail.ts
+++ b/web/src/components/pages/agent-detail.ts
@@ -909,9 +909,7 @@ export class ScionPageAgentDetail extends LitElement {
       <sl-tab-group @sl-tab-show=${this.handleTabShow}>
         <sl-tab slot="nav" panel="status">Status</sl-tab>
         ${this.agent.cloudLogging ? html`<sl-tab slot="nav" panel="logs">Logs</sl-tab>` : nothing}
-        ${this.agent.cloudLogging
-          ? html`<sl-tab slot="nav" panel="messages">Messages</sl-tab>`
-          : nothing}
+        <sl-tab slot="nav" panel="messages">Messages</sl-tab>
         <sl-tab slot="nav" panel="configuration">Configuration</sl-tab>
 
         <sl-tab-panel name="status">${this.renderStatusTab()}</sl-tab-panel>
@@ -930,17 +928,13 @@ export class ScionPageAgentDetail extends LitElement {
               </sl-tab-panel>
             `
           : nothing}
-        ${this.agent.cloudLogging
-          ? html`
-              <sl-tab-panel name="messages">
-                <scion-agent-message-viewer
-                  agentId=${this.agentId}
-                  agentName=${this.agent.name || ''}
-                  ?canSend=${can(this.agent._capabilities, 'message')}
-                ></scion-agent-message-viewer>
-              </sl-tab-panel>
-            `
-          : nothing}
+        <sl-tab-panel name="messages">
+          <scion-agent-message-viewer
+            agentId=${this.agentId}
+            agentName=${this.agent.name || ''}
+            ?canSend=${can(this.agent._capabilities, 'message')}
+          ></scion-agent-message-viewer>
+        </sl-tab-panel>
         <sl-tab-panel name="configuration">${this.renderConfigurationTab()}</sl-tab-panel>
       </sl-tab-group>
     `;
--- a/web/src/components/shared/agent-message-viewer.ts
+++ b/web/src/components/shared/agent-message-viewer.ts
@@ -74,6 +74,15 @@ export class ScionAgentMessageViewer extends LitElement {
   canSend = false;
 
   /**
+   * Whether Cloud Logging is available for this agent. When false, the
+   * viewer skips Cloud-Logging-only paths: the fallback fetch from
+   * /message-logs and the SSE stream from /message-logs/stream. The hub
+   * message store path still works and is the primary source regardless.
+   */
+  @property({ type: Boolean })
+  cloudLogging = false;
+
+  /**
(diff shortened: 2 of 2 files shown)


meatballs:
Good catch — addressed in `b4c4b00a`.

The viewer now takes a `cloudLogging` boolean prop (plumbed through from `agent-detail.ts`) and:

- **Skips the Cloud Logging fallback** in `fetchMessages()` when `cloudLogging` is false and no explicit `logsUrl` override is set. An empty hub-store result now renders the intended 'No messages found' empty state instead of falling through to a 501 error.
- **Hides the Stream toggle** in the toolbar when `cloudLogging` is false and no explicit `streamUrl` override is set. The streaming indicator markup is kept in case a future PR wires per-agent streaming to the existing `MessageBrokerProxy` event bus that inbox-tray already consumes.

Existing grove-detail usage of the viewer doesn't set the new prop, so it defaults to `false` — same correct behaviour on deployments without Cloud Logging. Agents that do have Cloud Logging enabled continue to get the fallback and stream as before.

ptone:
# PR #142 Review: fix(web): decouple Messages tab from cloudLogging gate

**Reviewer:** Automated Code Review
**Date:** 2026-04-13
**Files Changed:** `agent-detail.ts`, `agent-message-viewer.ts` (+38/-21)

---

## Executive Summary

This is a low-risk, well-scoped UI fix that removes an outdated feature gate so the Messages tab is visible regardless of Cloud Logging configuration. The changes are logically sound, with proper guarding of Cloud-Logging-only code paths in the viewer component, and no security or correctness issues were found.

---

## Critical Issues

None.

---

## Observations

### 1. Silent failure on Hub API errors when Cloud Logging is disabled (Low / Pre-existing)

**Location:** `agent-message-viewer.ts:406-422`

When `hubRes.ok` is `false` (e.g., hub returns 500) and `cloudLogging` is `false`, `fetchMessages()` now exits silently via the new guard at line 422 without surfacing an error. The `finally` block correctly sets `loading = false`, but no error message is shown to the user - they just see the empty state.

Previously this path was invisible because the entire tab was hidden for non-CL deployments, so this is not a regression from this PR. However, now that the tab is always visible, a transient hub failure will silently show "No messages found" instead of an error.

**Suggested improvement (non-blocking, follow-up):**
```typescript
// After line 416, before the CL fallback guard:
if (this.agentId && !hubRes.ok) {
  throw new Error(`Hub API returned HTTP ${hubRes.status}`);
}
```

### 2. `resolvedStreamUrl` still returns a URL when `cloudLogging` is false (Low / Cosmetic)

**Location:** `agent-message-viewer.ts:394-397`

The `resolvedStreamUrl` getter unconditionally returns the `/message-logs/stream` URL if `agentId` is set, regardless of the `cloudLogging` flag. This is safe because the UI hides the Stream toggle when `!streamAvailable`, so users can't trigger `startStream()` through the UI. However, if `startStream()` were ever called programmatically (e.g., from a parent component via the public API), it would attempt to connect to a 501 endpoint.

No action needed now - the UI guard is sufficient, and this would only matter if the component API changes.

### 3. Guard logic correctness: `!this.cloudLogging && !this.logsUrl`

**Location:** `agent-message-viewer.ts:422`

The guard `if (!this.cloudLogging && !this.logsUrl) return;` correctly preserves the existing behavior for consumers that pass a custom `logsUrl` (e.g., the inbox tray or test harnesses) while skipping the Cloud-Logging-only endpoint for standard non-CL deployments. This is well-thought-out.

### 4. `?cloudLogging=${this.agent.cloudLogging || false}` — unnecessary fallback

**Location:** `agent-detail.ts` (new line in template)

The `|| false` is technically redundant since Lit's `?` boolean attribute binding already treats `undefined`/`null` as `false`. Harmless but unnecessary.

---

## Positive Feedback

- **Clean separation of concerns:** Pushing the `cloudLogging` awareness down into the viewer component (rather than gating the entire tab) is the correct architectural choice. The viewer can now degrade gracefully per-feature rather than being an all-or-nothing gate.

- **Stream toggle guarding is well-handled:** Conditionally rendering the Stream toggle via `streamAvailable` (line 768) is the right UX — hiding a control that would silently fail is better than showing it and catching errors.

- **`return` inside `try` still hits `finally`:** The early return at line 422 correctly triggers the `finally { this.loading = false }` block, so no loading spinner gets stuck. This is a common gotcha and was handled correctly.

- **Thorough PR description:** The body explains the provenance of the original gate, why it became stale, the known limitation (stream endpoint), and the test matrix. Exemplary PR hygiene.

---

## Final Verdict

**Approve.** The changes are correct, minimal, and well-documented. The one pre-existing gap (silent hub errors) is worth tracking as a follow-up but is not a blocker for this PR.

PR #145, opened 2026-04-12
[Feature] minimal hub container image

### LINKED ISSUE #133: Helm chart and minimal hub/broker container image for Kubernetes deployment
## Problem / Motivation

I want to deploy to kubernetes, I will produce an helm chart no matter what, to make sure everyone don't have todo it I can spend the time building one and make it something everyone can use.

I understand if you think it's to early to have to maintain a helm chart, but it would also help with on-boarding of more users to the project. The main maintaince burdon would probably adding args to the entrypoint, but we can make that very general in the helm chart so it's up to the user to add any config they want witout extra validation from the chart.

For production Kubernetes deployments, users need:
1. A minimal container image containing the `scion` binary + embedded web assets (the existing agent images in `image-build/` are designed for agent containers, not the hub server)
2. A Helm chart that deploys the hub/broker with proper configuration, secrets, persistence, health checks, and RBAC

Of course the container and helm chart would be sepreate PRs

### 1. Minimal Hub/Broker Container Image

Build and publish a minimal container image for the `scion` server binary using [ko](https://ko.build/). Since the binary is already pure Go (`CGO_ENABLED=0`), ko is a natural fit:

- Produces minimal, distroless-based images with no shell or package manager
- Handles multi-arch (amd64/arm64) natively
- Integrates with existing GitHub Actions release workflow
- Makes SBOM generation and image signing straightforward to add later (cosign/syft)

The image only needs the `scion` binary with embedded web assets (built with `go build ./cmd/scion` — same binary serves hub, broker, and CLI). Entrypoint would be the `scion` binary directly.

**Publish to:** `ghcr.io/googlecloudplatform/scion` (or `ghcr.io/googlecloudplatform/scion-server` to distinguish from agent images)

**Trigger:** On tag push (alongside the existing `build-release.yml` workflow)

### 2. Helm Chart

Provide an helm chart as part of the release flow, It would be built using OCI so no index.yaml needed....

### Architecture context

Current constraint: SQLite limits the hub to a single replica. This is a known temporary limitation. [PostgreSQL support is planned](https://github.com/GoogleCloudPlatform/scion/issues/53), which would make the hub stateless and enable multi-replica scaling. The `EventPublisher` interface already has a designed path to `PostgresEventPublisher` using `LISTEN/NOTIFY` (see `.design/hosted/web-realtime.md`, Stage 5).

## Alternatives Considered

- **Kustomize**: Simpler but less flexible for the configuration surface area needed (two modes, many optional features). Helm's templating is better suited here.
- **Dockerfile instead of ko**: ko is preferred because the binary is pure Go (`CGO_ENABLED=0`), ko produces smaller/more secure images, and it makes multi-arch, SBOM, and signing trivial to add.

## Additional Context

- Design doc for event scaling path: `.design/hosted/web-realtime.md`

nissessenap:
Add ko configuration and a GitHub Actions workflow to build, publish, and sign a multi-arch (amd64/arm64) hub container image on every tag push. The image uses a distroless base (cgr.dev/chainguard/static) and embeds the web frontend assets.

The image is published as `ghcr.io/<owner>/scion:<tag>` and tagged with both the version tag and `latest`.

Closes #133 (container image portion)

- [X] Tests pass
- [X] Appropriate changes to documentation are included in the PR

### DIFF
--- a/.github/workflows/build-container.yml
+++ b/.github/workflows/build-container.yml
@@ -1,0 +1,130 @@
+# Copyright 2026 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: Build Hub Container Image
+
+on:
+  push:
+    tags:
+      - "*"
+
+jobs:
+  build:
+    name: Build and Push Hub Image
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+      id-token: write
+    outputs:
+      digest: ${{ steps.build-image.outputs.digest }}
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+          persist-credentials: false
+
+      - name: Set up Go
+        uses: actions/setup-go@v6
+        with:
+          go-version-file: "go.mod"
+          cache: true
+
+      - name: Setup Node
+        uses: actions/setup-node@v6
+        with:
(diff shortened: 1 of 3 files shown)


nissessenap:
Could argue we should updat docs to say that we provide a production based image for hub

ptone:
ko image + Helm chart, both superseded;

PR #66, opened 2026-04-08
feat: add OAuth token auth support for Claude harness

donovan-yohan:
## Summary

Closes #64

- Add `CLAUDE_CODE_OAUTH_TOKEN` env var support to `GatherAuthWithEnv`
- Add `ClaudeOAuthToken` field to `AuthConfig` and `OAuthToken` capability to `HarnessAuthCapabilities`
- Implement `oauth-token` auth method in Claude harness `ResolveAuth` with auto-detect priority: API key → OAuth token → Vertex AI
- Add `DetectAuthTypeFromEnvVars` support for the new auth type
- Add `RequiredAuthEnvKeys` entry for `oauth-token`
- Include debug logging for the new credential field
- Add comprehensive tests: auto-detect, explicit selection, missing token error, priority ordering (API key > OAuth > Vertex), and updated no-creds error message

## Test plan

- [ ] `make test` passes — all new and existing auth tests green
- [ ] Set `CLAUDE_CODE_OAUTH_TOKEN` and verify agent starts with `oauth-token` method
- [ ] Confirm API key still takes priority when both are set
- [ ] Confirm Vertex AI still works when OAuth token is absent

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #149, opened 2026-04-13
feat(harness): add OAuth token and credentials-file auth for Claude

### LINKED ISSUE #64: Add OAuth token auth support for Claude harness
## Problem / Motivation

Currently the Claude harness only supports two auth methods: direct API key (`ANTHROPIC_API_KEY`) and Vertex AI (ADC + project/region). Users with Claude Pro/Team/Enterprise subscriptions can generate OAuth tokens via `claude setup-token`, but there's no way to pass these tokens to containerized agents.

This means subscription users must either pay separately for API credits or set up Vertex AI, even though they already have a working auth method on their host machine.

## Proposed Solution

Add `oauth-token` as a third auth method for the Claude harness:

- **New env var**: `CLAUDE_CODE_OAUTH_TOKEN` — gathered from the environment alongside existing auth vars.
- **New field**: `ClaudeOAuthToken` on `AuthConfig` and `OAuthToken` on `HarnessAuthCapabilities`.
- **Auto-detect priority**: API key → OAuth token → Vertex AI (OAuth slots in between existing methods).
- **Explicit selection**: `auth type: oauth-token` in agent config, with a clear error if the token is missing.
- **Auth detection**: `DetectAuthTypeFromEnvVars` returns `"oauth-token"` when `CLAUDE_CODE_OAUTH_TOKEN` is present for the claude harness.

The implementation includes tests for: auto-detection, explicit selection, missing token errors, priority ordering (API key wins over OAuth, OAuth wins over Vertex), and updated no-creds error messaging.

## Alternatives Considered

- Reading the token file (`~/.claude/oauth_token`) directly from the container filesystem — but this couples the harness to Claude Code's internal file layout and doesn't work well across container boundaries. An env var is more portable.

## Additional Context

Token generation: `claude setup-token` (available in Claude Code CLI).

meatballs:
Closes #64.

## Summary

Adds two new auth methods to the Claude harness so Claude subscription users can run SCION agents without an Anthropic API key or Vertex AI setup:

1. **`oauth-token`** — long-lived token from `claude setup-token`, passed via the `CLAUDE_CODE_OAUTH_TOKEN` env var. Safe to share across concurrent agents; equivalent in scope to API-key auth (inference only).
2. **`auth-file`** — projects `~/.claude/.credentials.json` into the container as an opaque file via a `CLAUDE_AUTH` hub file secret, matching the existing pattern for Gemini OAuth, Codex, and OpenCode. Claude Code reads and refreshes the file natively inside the container, enabling full-scope subscription features like Remote Control that the inference-only token cannot access.

## Credit and context

This PR stands on the shoulders of #66 by @donovan-yohan, which opened this conversation and implemented the `oauth-token` half end-to-end. Thanks for getting the ball rolling — the env-var plumbing here follows the same shape and wouldn't have been this quick to put together without it.

Filing this as a new PR rather than pushing changes into #66 because working through the implementation end-to-end led to a meaningfully different approach on the credentials-file path (see below). Keeping the two approaches in separate PRs lets them be reviewed on their merits side-by-side.

## How this differs from #66

#66's first commit adds `CLAUDE_CODE_OAUTH_TOKEN` env-var support. This PR keeps that approach verbatim — it's the right solution for users who want a stable long-lived token and don't need Remote Control.

#66's second commit (`feat: add implicit auth auto-detection for Claude credentials file`) reads `~/.claude/.credentials.json` on the host, extracts the `accessToken` field, and funnels it through the same `CLAUDE_CODE_OAUTH_TOKEN` env-var path. This PR takes a different approach on that half, for three reasons:

1. **The access token in `.credentials.json` rotates.** Claude Code refreshes it every few hours using the refresh token, and the file is rewritten in place. Scraping the access token at gather time hands the container a snapshot that will expire mid-session, with no refresh path (the container can't write back to the host file). Long-running agents would silently start 401'ing when the snapshot ages out.

2. **Full-scope subscription features need the file, not a token.** Remote Control in particular is gated on the credentials file being present: tokens delivered via `CLAUDE_CODE_OAUTH_TOKEN` are treated as inference-only regardless of their origin. Claude Code prefers the env var over the file when both are set, so scraping the token and re-injecting it as env silently downgrades auth-file users back to the inference-only path — `/remote-control` never appears inside the agent session even when a valid credentials file is reachable.

3. **The issue's own "Alternatives Considered" section rules this approach out.** From #64:
   > *Reading the token file (`~/.claude/oauth_token`) directly from the container filesystem — but this couples the harness to Claude Code's internal file layout and doesn't work well across container boundaries. An env var is more portable.*

Instead, this PR treats `.credentials.json` as an opaque file and projects it into the container via SCION's existing file-secret machinery — the same pattern @ptone pointed at in his comment on #64 (*"the general 'harness credential file' that is in other harness types"*). Slotting into the established precedent:

| Harness             | Hub file secret name | Container mount path                |
| ------------------- | -------------------- | ----------------------------------- |
| gemini              | `GEMINI_OAUTH_CREDS` | `~/.gemini/oauth_creds.json`        |
| codex               | `CODEX_AUTH`         | `~/.codex/auth.json`                |
| opencode            | `OPENCODE_AUTH`      | `~/.local/share/opencode/auth.json` |
| **claude** (new)    | **`CLAUDE_AUTH`**    | **`~/.claude/.credentials.json`**   |

SCION itself never parses the file — it's handed to Claude Code intact, and Claude Code's own refresh loop runs inside the container against its own copy. This also means `auth-file` scales cleanly across multiple concurrent agents: each one gets its own snapshot from the hub at start time, refreshes are container-local, and there's no host-side rotation race.

The two methods (`oauth-token` and `auth-file`) are deliberately distinct. Auto-detect priority is:

    api-key → oauth-token → auth-file → vertex-ai → error

API key wins over both subscription paths because it's the most stable. `oauth-token` wins over `auth-file` because the long-lived setup-token doesn't rotate underneath a long-running session. `auth-file` is auto-detected in non-broker mode by `stat`'ing `~/.claude/.credentials.json` (never read or parsed); in broker mode, it's selected when a `CLAUDE_AUTH` hub file secret is present.

## What's in the PR

Three commits:

1. **`feat(harness): add oauth-token and auth-file support for Claude`** — the main change. Adds the new auth fields to `api.AuthConfig`, the `OAuthToken` field to `HarnessAuthCapabilities`, the `oauth-token`/`auth-file` branches in Claude's `ResolveAuth`, the detection rules in `DetectAuthTypeFromEnvVars` / `DetectAuthTypeFromFileSecrets`, the `CLAUDE_AUTH` mapping in `OverlayFileSecrets`, the `CLAUDE_CODE_OAUTH_TOKEN` entry in `isAuthEnvKey`, the `claude-auth` kind in `authFileKind`, and the `RequiredAuthEnvKeys` entries for the new methods. Also wires the new auth type through every place that enumerates the valid set: the hub validator in `pkg/hub/harness_capabilities.go` (previously fell through to "Unknown auth type"), the `agent-v1`/`settings-v1` JSON schemas, the dashboard type definitions and dropdowns in `web/src/`, and the CLI validators + flag help in `cmd/`. Tests cover gather (env + file discovery), overlay (hub file secrets), broker-mode isolation, `ResolveAuth` (explicit selection, auto-detect, missing-creds errors, priority ordering, unknown-type error message), and capability defaults.

2. **`fix(agent): drop auth-candidate env keys from opts.Env after auth resolution`** — a real correctness bug surfaced while validating the above. `filterResolvedSecretsForResolvedAuth` correctly filtered `opts.ResolvedSecrets`, but `pkg/hub/httpdispatcher.go` pre-merges environment-type secrets into the dispatch request's `ResolvedEnv` map before the broker ever sees them, and `start_context.go`'s `BuildStartContext` copies `ResolvedEnv` straight into `opts.Env`. By the time the filter runs, any auth env secret the hub already pre-merged is sitting in `opts.Env` and untouched. Selecting `auth-file` for Claude would therefore still inject a stale `CLAUDE_CODE_OAUTH_TOKEN` env var alongside the mounted credentials file, and Claude Code's env-over-file preference would silently defeat the whole point. Mirroring the filter onto `opts.Env` after auth resolution closes the gap — and tightens auth isolation for `vertex-ai` and `api-key` on all harnesses as a side benefit.

3. **`test(runtimebroker): isolate HOME in env-gather test helpers`** — a pre-existing test fragility, surfaced while landing the rest of this work. `LoadEffectiveSettings` merges the developer's personal `~/.scion/settings.yaml` on top of the test grove's settings, so a developer with a user-scoped `harness_configs.<harness>.auth_selected_type` would see env-gather tests resolve their personal auth type instead of the fixture's. `main` hides this by accident (unknown auth types fall through to nil required-env-keys, so the tests return 201 anyway), but real support for `oauth-token` surfaces it. `t.Setenv("HOME", t.TempDir())` in both test helpers fixes it and, as a side effect, also fixes the pre-existing `TestEnvGather_AutoDetectVertexAI_FromGACEnvVar` flake on dev machines with user-level Claude auth settings.

## Test plan

- [x] `make test` — new tests green; no new failures on touched packages.
- [x] Deployed end-to-end against a real hub + broker + containerised Claude Code agent:
  - [x] Existing `oauth-token` agents continue to authenticate and round-trip messages unchanged against an existing `CLAUDE_CODE_OAUTH_TOKEN` hub secret. Filter correctly keeps the secret when `oauth-token` is the selected method.
  - [x] `auth-file` agent with `CLAUDE_AUTH` hub file secret:
    - `~/.claude/.credentials.json` is present inside the container (owned by the container user, mode 0600).
    - `CLAUDE_CODE_OAUTH_TOKEN` is correctly absent from the container env (filter working).
    - Claude Code banner shows "Claude Max" — full-scope subscription auth, not inference-only.
    - `/remote-control` slash command is available inside the agent session for the first time.
  - [x] Dashboard agent-create and agent-configure pages show the new "OAuth Token" dropdown option and validate correctly against the hub.

### DIFF
--- a/cmd/common.go
+++ b/cmd/common.go
@@ -321,10 +321,10 @@ func RunAgent(cmd *cobra.Command, args []string, resume bool) error {
 	// Validate --harness-auth value
 	if harnessAuthFlag != "" {
 		switch harnessAuthFlag {
-		case "api-key", "vertex-ai", "auth-file":
+		case "api-key", "oauth-token", "auth-file", "vertex-ai":
 			// valid
 		default:
-			return fmt.Errorf("invalid --harness-auth value %q: must be one of api-key, vertex-ai, auth-file", harnessAuthFlag)
+			return fmt.Errorf("invalid --harness-auth value %q: must be one of api-key, oauth-token, auth-file, vertex-ai", harnessAuthFlag)
 		}
 	}
 
--- a/cmd/config_schema.go
+++ b/cmd/config_schema.go
@@ -66,7 +66,7 @@ func configSchemaFields() []schemaField {
 		{Name: "image", Type: "string", Description: "Container image"},
 		{Name: "model", Type: "string", Description: "LLM model to use"},
 		{Name: "user", Type: "string", Description: "Container unix user"},
-		{Name: "auth_selected_type", Type: "string", Description: "Auth method (api-key, vertex-ai, auth-file)"},
+		{Name: "auth_selected_type", Type: "string", Description: "Auth method (api-key, oauth-token, auth-file, vertex-ai)"},
 		{Name: "task", Type: "string", Description: "Task prompt for the agent"},
 		{Name: "branch", Type: "string", Description: "Git branch for the agent workspace"},
 		{Name: "system_prompt", Type: "string", Description: "System prompt (inline text or file:// URI)"},
--- a/cmd/create.go
+++ b/cmd/create.go
@@ -50,10 +50,10 @@ arguments are provided, an empty prompt.md is created for later editing.`,
 		// Validate --harness-auth value
 		if harnessAuthFlag != "" {
 			switch harnessAuthFlag {
-			case "api-key", "vertex-ai", "auth-file":
+			case "api-key", "oauth-token", "auth-file", "vertex-ai":
 				// valid
 			default:
-				return fmt.Errorf("invalid --harness-auth value %q: must be one of api-key, vertex-ai, auth-file", harnessAuthFlag)
+				return fmt.Errorf("invalid --harness-auth value %q: must be one of api-key, oauth-token, auth-file, vertex-ai", harnessAuthFlag)
 			}
 		}
 
@@ -282,7 +282,7 @@ func init() {
 	createCmd.Flags().StringVar(&harnessConfigFlag, "harness-config", "", "Named harness configuration to use")
 	createCmd.Flags().StringVar(&harnessConfigFlag, "harness", "h", "Named harness configuration to use (alias for --harness-config)")
 
-	createCmd.Flags().StringVar(&harnessAuthFlag, "harness-auth", "", "Override auth method for the harness (api-key, vertex-ai, auth-file)")
+	createCmd.Flags().StringVar(&harnessAuthFlag, "harness-auth", "", "Override auth method for the harness (api-key, oauth-token, auth-file, vertex-ai)")
 
 	// Template resolution flags for Hub mode (Section 9.4)
 	createCmd.Flags().BoolVar(&uploadTemplate, "upload-template", false, "Automatically upload local template to Hub if not found")
--- a/cmd/start.go
+++ b/cmd/start.go
@@ -56,7 +56,7 @@ func init() {
 	startCmd.Flags().StringVar(&harnessConfigFlag, "harness-config", "", "Named harness configuration to use")
 	startCmd.Flags().StringVar(&harnessConfigFlag, "harness", "", "Named harness configuration to use (alias for --harness-config)")
 
-	startCmd.Flags().StringVar(&harnessAuthFlag, "harness-auth", "", "Override auth method for the harness (api-key, vertex-ai, auth-file)")
+	startCmd.Flags().StringVar(&harnessAuthFlag, "harness-auth", "", "Override auth method for the harness (api-key, oauth-token, auth-file, vertex-ai)")
 
 	// Notification flag
 	startCmd.Flags().BoolVar(&notify, "notify", false, "Get notified when the spawned agent reaches a terminal state")
--- a/pkg/agent/run.go
+++ b/pkg/agent/run.go
@@ -363,13 +363,15 @@ func (m *AgentManager) Start(ctx context.Context, opts api.StartOptions) (*api.A
 		if opts.BrokerMode {
 			harness.OverlayFileSecrets(&auth, opts.ResolvedSecrets)
 		}
-		util.Debugf("auth: gathered credentials — selectedType=%q, hasGeminiKey=%t, hasGoogleKey=%t, hasOAuth=%t, hasADC=%t, hasAnthropicKey=%t, cloudProject=%q, gcpMetadataMode=%q, brokerMode=%t",
+		util.Debugf("auth: gathered credentials — selectedType=%q, hasGeminiKey=%t, hasGoogleKey=%t, hasOAuth=%t, hasADC=%t, hasAnthropicKey=%t, hasClaudeOAuthToken=%t, hasClaudeAuthFile=%t, cloudProject=%q, gcpMetadataMode=%q, brokerMode=%t",
 			auth.SelectedType,
 			auth.GeminiAPIKey != "",
 			auth.GoogleAPIKey != "",
 			auth.OAuthCreds != "",
 			auth.GoogleAppCredentials != "",
 			auth.AnthropicAPIKey != "",
+			auth.ClaudeOAuthToken != "",
+			auth.ClaudeAuthFile != "",
 			auth.GoogleCloudProject,
 			auth.GCPMetadataMode,
 			opts.BrokerMode,
@@ -1064,6 +1066,7 @@ func isAuthEnvKey(key string) bool {
 	case "GEMINI_API_KEY",
 		"GOOGLE_API_KEY",
 		"ANTHROPIC_API_KEY",
+		"CLAUDE_CODE_OAUTH_TOKEN",
 		"OPENAI_API_KEY",
 		"CODEX_API_KEY",
 		"GOOGLE_CLOUD_PROJECT",
@@ -1088,6 +1091,8 @@ func authFileKind(name, target string) string {
 		return "codex-auth"
 	case name == "OPENCODE_AUTH" || strings.HasSuffix(target, "/opencode/auth.json"):
 		return "opencode-auth"
+	case name == "CLAUDE_AUTH" || strings.HasSuffix(target, "/.claude/.credentials.json"):
+		return "claude-auth"
 	default:
 		return ""
 	}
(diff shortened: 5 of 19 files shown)


ptone:
# PR #149 Review: feat(harness): add OAuth token and credentials-file auth for Claude

## Executive Summary

This PR adds two well-motivated auth methods (`oauth-token` and `auth-file`) for Claude subscription users, with a solid correctness fix for env-key leakage in broker mode. The change is **low-to-medium risk** — the new code paths are clean, well-tested, and follow established patterns from other harnesses (Gemini, Codex, OpenCode). No blocking issues identified.

---

## Critical Issues

**None.**

---

## Observations

### 1. `agent-create.ts` — missing `?disabled` guard on `oauth-token` option

**File:** `web/src/components/pages/agent-create.ts:1160`

On the **configure** page, the new `oauth-token` option correctly checks `?disabled=${this.isUnsupported(oauthTokenCap)}`, but on the **create** page, no capability-based disable is applied to any auth option (including `oauth-token`). This is a **pre-existing inconsistency** — vertex-ai and auth-file also lack the guard on the create page — but it means a user could select `oauth-token` during creation for a harness that doesn't support it. Server-side validation in `validateConfigAgainstHarnessCapabilities` will catch it, so there's no security concern, but it's a UX gap.

**Severity:** Low (pre-existing pattern, not introduced by this PR, server rejects invalid combos).

### 2. `opts.Env` filtering — correctness of `resolvedForSecretFilter`

**File:** `pkg/agent/run.go:416-429`

The new env-key filter builds `requiredAuthEnv` from `resolvedForSecretFilter.EnvVars`. This is correct because `resolvedForSecretFilter` mirrors the resolved auth struct and contains only the env keys the chosen auth method actually needs. The `delete(opts.Env, k)` call during map iteration is safe in Go (range over map handles concurrent deletion). Logic is sound.

### 3. `ContainerPath` uses `~` tilde

**File:** `pkg/harness/claude_code.go:514,557`

```go
{SourcePath: auth.ClaudeAuthFile, ContainerPath: "~/.claude/.credentials.json"},
```

The `ContainerPath` uses `~` which requires downstream tilde expansion. This follows the exact same pattern used by Gemini (`~/.gemini/oauth_creds.json`), Codex (`~/.codex/auth.json`), and OpenCode (`~/.local/share/opencode/auth.json`) in their respective harness implementations, so this is consistent with the existing contract.

**Severity:** None (follows established convention).

### 4. `DetectAuthTypeFromFileSecrets` — broker-mode priority

The comment says `api-key -> oauth-token (env) -> auth-file -> ADC (vertex-ai)`. The file-secret detection correctly checks `CLAUDE_AUTH` before `gcloud-adc`. The env-var detection correctly checks `CLAUDE_CODE_OAUTH_TOKEN` before `GOOGLE_APPLICATION_CREDENTIALS`. These two detection functions are called from separate code paths in the env-gather handler, and the caller is responsible for combining them with the correct precedence. The ordering within each function is correct and mirrors the documented priority.

### 5. `RequiredAuthEnvKeys` returns `nil` for `auth-file`

**File:** `pkg/harness/auth.go:341`

Returning `nil` for `auth-file` means the env-gather endpoint won't require any environment secrets for this auth method — correct, since auth-file relies on a file mount, not env vars. This matches the pattern for other harnesses' `auth-file` methods.

### 6. Minor: Debug log format string growth

**File:** `pkg/agent/run.go:366`

The debug log line now has 11 format arguments. Not a problem, but if this pattern continues to grow, consider using structured logging (e.g., a map). Not blocking.

---

## Positive Feedback

1. **Excellent test coverage.** The PR adds tests for: auto-detect priority ordering, explicit selection with missing credentials, unknown type error messages, file discovery, broker-mode isolation, and overlay file secrets. The priority test (`TestClaudeResolveAuth_Priority`) systematically validates each tier of the fallback chain — this is exactly the kind of test that prevents regressions.

2. **The `opts.Env` filter fix (commit 2) is a real catch.** The explanation of how `httpdispatcher.go` pre-merges env secrets before the broker sees them, creating a gap in `filterResolvedSecretsForResolvedAuth`, is convincing. This fixes a subtle auth-isolation bug that would have caused `CLAUDE_CODE_OAUTH_TOKEN` to silently override the credentials-file mount via Claude Code's env-over-file preference.

3. **Opaque file treatment is the right design.** Not parsing `~/.claude/.credentials.json` and instead mounting it as-is avoids both token-rotation staleness and coupling to Claude Code's internal file format. The comments explaining this decision are clear and well-placed.

4. **Test isolation fix (commit 3) is thoughtful.** `t.Setenv("HOME", t.TempDir())` prevents developer-local settings from contaminating test results — the kind of fix that prevents future flaky CI runs.

5. **Consistent cross-cutting changes.** CLI validators, JSON schemas, hub capabilities, dashboard dropdowns, and type definitions are all updated in lockstep. No gaps in the enumeration of valid auth types across the codebase.

---

## Final Verdict

**Approve.** The implementation is clean, well-tested, and follows established patterns. The auth-isolation fix in commit 2 adds tangible correctness value beyond the feature itself. No blocking issues found.

PR #151, opened 2026-04-13
feat(hub): add native discord notification channel

meatballs:
## Problem

`pkg/hub/channels_slack.go` validates webhook URLs very loosely — any `https://` URL that the hub can POST to is accepted — because Discord-using operators route notifications through Discord's Slack-compat endpoint by appending `/slack` to the webhook URL. That accommodation makes Slack validation meaningfully weaker than it should be: an operator who typos a URL, pastes an unrelated webhook, or is targeted by a config-injection attack has no safety net from the channel factory.

## Solution

Add a dedicated `discord` notification channel type that speaks the native Discord webhook format (rich embeds, colour-coded severity, `allowed_mentions`-controlled role mentions) and strictly validates its webhook URLs against the four allowed Discord hostnames (`discord.com`, `discordapp.com`, `ptb.discord.com`, `canary.discord.com`), explicitly rejecting `/slack` suffixes.

With a first-class Discord channel available, the Slack channel can eventually have its URL validation tightened back to `hooks.slack.com` only (out of scope for this PR — happy to file as a follow-up).

### Features

- **Native webhook format**: embeds with title, description, colour, and RFC 3339 timestamp — no Slack-compat shim
- **Colour mapping**: `state-change → blue`, `input-needed → yellow`, `instruction → grey`, and `Urgent: true → red` (overrides the type colour)
- **Strict URL validation**: rejects non-Discord hosts and explicitly rejects `/slack`-suffixed URLs with a descriptive error, so the operator intent is clear
- **Embed-length truncation**: honours Discord's 2048-char description cap and 6000-char total limit, rune-safe, with a trailing ellipsis marker
- **Safe mentions**: `allowed_mentions.parse: []` is always set, so literal `@everyone`/`@here` in message bodies never trigger mass pings; role/user IDs are opt-in via an explicit `mention_on_urgent` param in Discord's `<@&ROLE>` / `<@USER>` syntax
- **Filters**: inherits the registry's `filter_types` and `filter_urgent_only` support — no channel-specific logic

### Example config

```yaml
notification_channels:
  - type: discord
    params:
      webhook_url: https://discord.com/api/webhooks/123/abc
      mention_on_urgent: "<@&987654321>"  # optional role ping on Urgent: true
      username: "scion"                   # optional webhook username override
      avatar_url: "https://example.com/scion.png"  # optional avatar
```

## Production evidence

Running on `scion.tanti.org.uk` (our deployed hub) since **2026-04-13**. Live-verified against a real Discord channel:

- State-change notifications render as blue embeds
- Input-needed notifications (via an `AskUserQuestion` trigger inside the Claude Code harness) render as yellow embeds
- Urgent-red path covered by unit tests (`TestDiscordChannel_Formatting`)
- `@everyone` / `@here` literal strings in message bodies do NOT trigger mass pings — `allowed_mentions.parse: []` holding as designed

## Scope

**In scope**
- `pkg/hub/channels_discord.go`: `DiscordChannel` type, `formatDiscordPayload` builder, URL validator, colour mapper, truncation helper
- `pkg/hub/channels.go`: `"discord"` case in `newChannelFromConfig`
- `pkg/hub/channels_test.go`: `TestDiscordChannel_*` suite
- `docs-site/src/content/docs/reference/server-config.md`: user-facing reference docs for all notification channel types, with the Discord subsection as the primary addition and a migration note for operators currently using the `/slack`-suffix hack

**Out of scope** (happy to file as follow-ups if wanted)
- A Discord bot gateway for receiving replies — this PR is outbound-only
- Slash commands or interactive components
- Tightening Slack URL validation back to `hooks.slack.com` — will break existing `/slack`-routed Discord setups until they migrate, so should land in a separate PR with a deprecation cycle
- A notification-channel filter that skips non-enum message types (a related finding from the production rollout — filed separately)

## Testing

All tests run under `make ci` / `go test ./...`:

- URL validation: all four Discord hosts accepted, `/slack`-suffixed URLs rejected, non-Discord hosts rejected
- Payload shape: embed title, description, colour, timestamp, `allowed_mentions` lockdown
- Colour mapping per message type (blue/yellow/grey) + urgent override (red)
- Truncation at 2048 chars per description with ellipsis marker, rune-safe
- Urgent role mention renders via `allowed_mentions.roles` and never triggers `@everyone`/`@here`
- Registry wiring: `newChannelFromConfig("discord", …)` returns `*DiscordChannel`

### DIFF
--- a/docs-site/src/content/docs/reference/server-config.md
+++ b/docs-site/src/content/docs/reference/server-config.md
@@ -179,3 +179,161 @@ When `server.hub.public_url` is not explicitly set, the Hub endpoint injected in
 4. Auto-computed `http://localhost:{port}` (last resort).
 
 For local development where the Hub runs on `localhost` but agents are in containers, set `server.broker.container_hub_endpoint` to a container-accessible address like `http://host.containers.internal:8080`.
+
+## Notification channels
+
+Notification channels deliver agent messages to external systems. Configure them
+under `server.hub.notification_channels` as a list of channel objects. Each object
+has a `type`, a `params` map, and optional filters.
+
+```yaml
+server:
+  hub:
+    notification_channels:
+      - type: <channel-type>
+        params:
+          # channel-specific key/value pairs
+        filter_urgent_only: false   # if true, only deliver urgent messages
+        filter_types:               # if set, only deliver these message types
+          - input-needed
+          - state-change
+```
+
+### Slack channel
+
+Delivers notifications via a Slack incoming webhook using Slack's `text` payload
+format.
+
+**Type:** `slack`
+
+**Parameters:**
+
+| Param              | Required | Description |
+|--------------------|----------|-------------|
+| `webhook_url`      | yes      | Slack incoming webhook URL (must use `https://`). |
+| `channel`          | no       | Override the webhook's default channel. |
+| `mention_on_urgent`| no       | Mention string added when `msg.Urgent == true` (e.g. `@here`, `@channel`). |
+
+**Example:**
+
+```yaml
+notification_channels:
+  - type: slack
+    params:
+      webhook_url: https://hooks.slack.com/services/T00000000/B00000000/XXXXXXXX
+      mention_on_urgent: "@here"
+```
+
+### Webhook channel
+
+Delivers notifications as a raw HTTP POST to an arbitrary URL. Use this when you
+need the full structured payload without truncation or when integrating with a
+custom receiver.
+
+**Type:** `webhook`
+
+**Parameters:**
+
+| Param         | Required | Description |
+|---------------|----------|-------------|
+| `webhook_url` | yes      | Destination URL (must use `https://`). |
+
+**Example:**
+
+```yaml
+notification_channels:
+  - type: webhook
+    params:
+      webhook_url: https://example.com/scion-notifications
+```
+
+### Email channel
+
+Delivers notifications by email.
+
+**Type:** `email`
+
+**Parameters:**
+
+| Param    | Required | Description |
+|----------|----------|-------------|
+| `to`     | yes      | Recipient email address. |
+| `from`   | no       | Sender address override. |
+| `smtp`   | no       | SMTP server host:port. |
+
+**Example:**
+
+```yaml
+notification_channels:
+  - type: email
+    params:
+      to: oncall@example.com
+```
+
+### Discord channel
+
+Delivers notifications via a Discord incoming webhook using Discord's native
+webhook format (rich embeds, colour-coded severity, allowed-mentions-controlled
+role/user pings). Unlike the Slack channel, the Discord channel targets the
+Discord-native endpoint — the `/slack`-compatibility suffix is explicitly
+rejected because it dilutes what each channel type means and silently hides
+the user's real intent.
+
+**Type:** `discord`
+
+**Parameters:**
+
+| Param               | Required | Description |
+|---------------------|----------|-------------|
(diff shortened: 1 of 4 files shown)


ptone:
Thanks for this - some of the message broker stuff is pretty new I want to give this a good look. Have you seen the pattern established in https://github.com/GoogleCloudPlatform/scion/tree/main/extras/scion-chat-app - should it go under that extra?

ptone:
see also: https://github.com/GoogleCloudPlatform/scion/blob/main/.design/slack-chat.md

meatballs:
Thanks for the pointers. I looked at both `extras/scion-chat-app` and `.design/slack-chat.md` carefully before replying, because I want to make sure I'm not missing the intended direction.

I think this PR sits at a different layer than the chat app, and I'd like to make the case for keeping it where it is.

**What this PR actually is.** It is a notification *sink*, a sibling to `channels_slack.go`, `channels_email.go`, and `channels_webhook.go`. One-way, fire-and-forget HTTP delivery triggered by the hub's existing notification registry. Around 600 lines, no new dependencies, no inbound events, no commands, no identity, no state. The Discord analogue of `channels_slack.go`, not of `internal/googlechat/`.

**What `extras/scion-chat-app` is.** A bidirectional chat integration: slash commands, `Card`/`Dialog`/`Widget` rendering, identity mapping, per-user subscriptions, SQLite state, and, critically, delivery to the hub over the broker plugin RPC rather than the notification registry. A Discord adapter built to the `chatapp.Messenger` interface would need Interactions-endpoint signature verification, slash command registration, component interactions, an App Home equivalent, and identity linking. That is a multi-phase project on the scale of the Slack adapter in `.design/slack-chat.md`, not a notification channel.

**Deployment footprint is the sticking point.** The chat app assumes a GCP-backed deployment: a GCP project with the Chat API and Secret Manager enabled, a service account, the hub's user-signing key stored in Secret Manager (the chat app mints impersonation tokens from it), Workspace Add-on registration in the Cloud Console, and Cloud Run or a GCE VM with an attached service account. A starter-hub deployment on a plain VM has none of that. Provisioning a GCP footprint purely as scaffolding for Discord alerts would be a significant increase in operator burden, and I suspect most self-hosted starter-hub users would rather paste a webhook URL into `settings.yaml` and be done.

**The security motivation does not transfer.** The reason I wrote this was that `channels_slack.go` currently accepts any `https://` URL, because Discord operators route through Discord's `/slack` compatibility endpoint. That makes Slack validation meaningfully weaker than it should be. A first-class Discord notification channel lets Slack go back to `hooks.slack.com`-only validation (happy to file as a follow-up). Moving Discord into `extras/scion-chat-app` would not address that, because the Slack channel's lax validation would still have no replacement for Discord-webhook users who do not want to run the full chat app.

**How I would reconcile the two.** If Discord grows a full bidirectional integration later (slash commands, agent control, identity linking), that absolutely belongs in `extras/scion-chat-app` as a third `Messenger` alongside Google Chat and Slack, following the pattern in `slack-chat.md`. This PR does not preclude that; the two can coexist the same way a hypothetical Slack chat adapter would coexist with `channels_slack.go`. One is "hub to room alerts", the other is "room to agents".

Happy to add a note to the PR description, or a comment in `channels_discord.go`, explicitly marking this as the notification-path scope and pointing at `extras/scion-chat-app` as the home for a future bidirectional Discord integration, if that would help future readers find the right layer.

ptone:
thank you for taking the time to respond in detail with a sound POV. 

I'm fine to add this as a fire-and-forget option.  just know that the channels_slack.go, channels_email.go were completely unused to date, and in general that pattern should be something moved to the message broker pattern.  I'm OK to land this for now - especially if it is something you are actually using.

a full bidi discord app is absolutely on the TODO list here. and in general I think messaging generally may get some re-factoring. Ideally this framework uses extensibility to support N integrations, rather than accumulate too much integration specific bloat for each and every integration.

PR #173, opened 2026-04-19
Create log directory if it does not exist

HernandoR:
Ensure log directory is created before opening log file.

too small changes, don't think worth an separate issue

- [x] Tests pass
- [x] Appropriate changes to documentation are included in the PR


ptone:
Thank you for the contribution - the AI comment RE 700 is sound, but needs to be addressed more comprehensively in the codebase. For hosted systems, the local permission is not a large attack surface as interaction is via server/API - added small fixes.

PR #175, opened 2026-04-19
fix: render action buttons for agents added via realtime SSE

ptone:
SSE agent events don't carry _capabilities, so action buttons (start, stop, delete, terminal) were missing for newly created agents until page reload. The existing fallback to agentScopeCapabilities only worked when the API returned scope-level capabilities — when only per-agent capabilities were present, the scope field stayed undefined.

Three fixes:
- Derive agentScopeCapabilities from per-agent capabilities in loadData and fetchAndMergeAgents when scope-level capabilities are absent
- Lazily derive scope capabilities from existing agents during SSE update handling
- Preserve existing per-agent capabilities during SSE merges

- [ ] Tests pass

### DIFF
--- a/web/src/components/pages/grove-detail.ts
+++ b/web/src/components/pages/grove-detail.ts
@@ -613,14 +613,27 @@ export class ScionPageGroveDetail extends LitElement {
     const updatedAgents = stateManager.getAgents();
     // Merge SSE agent deltas into local agent list
     const agentMap = new Map(this.agents.map((a) => [a.id, a]));
+    // Lazily derive scope capabilities from existing agents if not yet set.
+    if (!this.agentScopeCapabilities) {
+      for (const a of agentMap.values()) {
+        if (a._capabilities) {
+          this.agentScopeCapabilities = a._capabilities;
+          break;
+        }
+      }
+    }
     for (const agent of updatedAgents) {
       if (agent.groveId === this.groveId || agentMap.has(agent.id)) {
         const existing = agentMap.get(agent.id);
         const merged = { ...existing, ...agent } as Agent;
         // New agents from SSE don't carry per-resource _capabilities.
         // Inherit scope-level capabilities so action buttons render.
-        if (!merged._capabilities && this.agentScopeCapabilities) {
-          merged._capabilities = this.agentScopeCapabilities;
+        if (!merged._capabilities) {
+          if (existing?._capabilities) {
+            merged._capabilities = existing._capabilities;
+          } else if (this.agentScopeCapabilities) {
+            merged._capabilities = this.agentScopeCapabilities;
+          }
         }
         agentMap.set(agent.id, merged);
       }
@@ -668,6 +681,11 @@ export class ScionPageGroveDetail extends LitElement {
         } else {
           this.agents = agentsData.agents || [];
           this.agentScopeCapabilities = agentsData._capabilities;
+        }
+        // Derive scope capabilities from per-agent capabilities when the
+        // response doesn't include a top-level _capabilities field.
+        if (!this.agentScopeCapabilities) {
+          this.agentScopeCapabilities = this.agents.find(a => a._capabilities)?._capabilities;
         }
       } else {
         // Fallback: if grove-scoped agents endpoint fails, try filtering from all agents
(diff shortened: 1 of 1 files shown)


ptone:
# PR #175 Review: fix: render action buttons for agents added via realtime SSE

**Author:** Preston Holmes (`ptone`)
**Branch:** `scion/missing-action-buttons` -> `main`
**Files changed:** 1 (`web/src/components/pages/grove-detail.ts`) — +23 / -2

---

## Executive Summary

This is a low-risk, well-scoped bug fix that addresses a real UX gap: action buttons (start, stop, delete, terminal) were missing for agents created via SSE because capabilities metadata was not being propagated. The fix is correct in its approach, with one minor semantic observation worth considering.

---

## Change Analysis

### What the PR does

The PR fixes three related paths where `_capabilities` was not available for agents:

1. **`onAgentsUpdated()` — SSE merge handler (lines 616-637):**
   - **New:** Lazily derives `agentScopeCapabilities` from existing agents' per-agent capabilities if the scope-level field was never set.
   - **Changed:** When merging an SSE agent delta that lacks `_capabilities`, the code now first tries to preserve the existing agent's `_capabilities` before falling back to `agentScopeCapabilities`. Previously, the spread merge (`{ ...existing, ...agent }`) would overwrite `existing._capabilities` with the SSE delta's `undefined` value.

2. **`loadData()` — initial data fetch (lines 687-689):**
   - After setting `agentScopeCapabilities` from the API response's top-level `_capabilities`, falls back to deriving it from the first agent that has per-agent capabilities.

3. **`fetchAndMergeAgents()` — background refresh (lines 744-746):**
   - Same fallback derivation as `loadData()`.

---

## Critical Issues

**None found.**

---

## Observations

### 1. Semantic mismatch: per-agent capabilities used as scope capabilities (Low Risk)

In `loadData()` and `fetchAndMergeAgents()`, when the API response lacks a top-level `_capabilities` field, the code derives `agentScopeCapabilities` from the first agent's `_capabilities`:

```typescript
this.agentScopeCapabilities = this.agents.find(a => a._capabilities)?._capabilities;
```

`agentScopeCapabilities` is used to gate scope-level UI controls like the **"Create Agent"** button (`can(this.agentScopeCapabilities, 'create')`) and **"Stop All"** button (`can(this.agentScopeCapabilities, 'stop_all')`). Per-agent capabilities (e.g., `attach`, `stop`, `start`, `delete`) may not contain scope-level actions like `create` or `stop_all`, meaning those buttons might still not render even after this fix.

However, this is a **best-effort improvement** over the status quo where these buttons were unconditionally hidden when `agentScopeCapabilities` was `undefined`. If per-agent capabilities happen to include scope actions (or if the backend returns uniform capabilities), this fix works correctly. If they don't, behavior is unchanged — no regression.

**Verdict:** Acceptable as-is. A future enhancement could have the backend always return scope capabilities, eliminating the need for this client-side derivation.

### 2. Object identity sharing (Informational)

The `_capabilities` object is shared by reference across agents:

```typescript
merged._capabilities = existing._capabilities;
// or
merged._capabilities = this.agentScopeCapabilities;
```

Since `Capabilities` is `{ actions: string[] }` and no code in this file mutates the `actions` array in-place, this is safe. Just noting that if a future change were to mutate capabilities in-place, it would silently affect all agents sharing the same reference. This is fine for now.

### 3. Spread merge correctly addressed (Positive)

The key insight of this PR is that `{ ...existing, ...agent }` will overwrite `existing._capabilities` with `undefined` when the SSE delta has `_capabilities: undefined` (since the key exists on the spread source). The fix correctly re-checks `existing._capabilities` after the merge, which is the right approach. An alternative would have been to strip `undefined` values from the SSE delta before spreading, but the current approach is more explicit and easier to reason about.

### 4. DRY opportunity (Minor)

The fallback derivation logic is duplicated between `loadData()` and `fetchAndMergeAgents()`:

```typescript
if (!this.agentScopeCapabilities) {
  this.agentScopeCapabilities = this.agents.find(a => a._capabilities)?._capabilities;
}
```

Consider extracting to a small private method like `ensureScopeCapabilities()` to avoid drift. Not blocking.

---

## Positive Feedback

- **Well-targeted fix** — all three code paths where capabilities could be lost are addressed.
- **Fail-closed security** — the `can()` function returns `false` when capabilities are `undefined`, so even if the fallback derivation fails, no action buttons are incorrectly shown. The worst case is buttons remaining hidden, not unauthorized actions being exposed.
- **Clear comments** — each change site has a comment explaining *why* the fallback exists.
- **Minimal blast radius** — 25 lines changed in a single file, all within the same logical concern.

---

## Final Verdict

**Approve.** This is a correct, minimal fix for a real UX bug. The semantic observation about scope-vs-agent capabilities is worth noting for the backend team, but does not block this PR. No security, correctness, or performance concerns.

PR #180, opened 2026-04-23
fix(broker): apply #159's whoami-skip-su pattern to remaining su call sites

jordantdavis:
## Summary

Following the pattern from #159, this extends the whoami-skip-su fix to the remaining `su -` call sites the original PR didn't cover, and factors the shared logic into a `runtime.ExecAsUserCmd` helper used by all six sites (the five being newly fixed plus #159's inline fix in `Attach`, opportunistically migrated to the helper for consistency).

## Background

PR #159 fixed `scion attach` on GKE Autopilot by checking `$(whoami)` before invoking `su - scion`, which prompts for a password when the container's `/etc/pam.d/su` lacks `pam_rootok.so` (the case on Debian slim bases like `node:20-slim` that the upstream `image-build/scion-base` chain extends). That fix landed only in `KubernetesRuntime.Attach`; the runtime-broker PTY paths and the K8s `Exec` wrapper kept the old shape.

## Symptom in cluster mode

Hub + runtime-broker on K8s, agents in separate pods: the dashboard's live terminal stream never connects. Hub logs show:

```
ERROR PTY stream error: timed out waiting for tmux session in container '<pod>' to become ready
```

…even though the tmux session **is** running fine inside the agent container. The agent works (Vertex calls succeed, harness responds, telemetry flows), but operators can't see the live PTY.

Verified on a GKE cluster:

```
$ kubectl exec <pod> -c agent -- ps -eo user,pid,comm
scion          1 sciontool
scion         78 tmux: server         ← session running fine
scion         79 claude

$ kubectl exec <pod> -c agent -- su - scion -c "tmux has-session -t scion"
Password: su: Authentication failure   ← the broker's exact probe command
```

The probe is the broker's `tmux has-session` check at `pkg/runtimebroker/pty_handlers.go:140`, which used the same `su - <user> -c ...` shape that #159 patched in the runtime package.

## Affected call sites

| File | Site | Action |
|---|---|---|
| `pkg/runtimebroker/pty_handlers.go` (4 sites) | window query, readiness probe, two PTY attach paths | **fixed** |
| `pkg/runtime/k8s_runtime.go` (Exec) | K8s exec wrapper | **fixed** |
| `pkg/runtime/k8s_runtime.go` (Attach) | the #159 site | **opportunistically migrated to the shared helper** |
| `pkg/runtime/k8s_podexec_test.go` | test asserting old `su -` command shape | updated to use the helper |
| `pkg/runtime/k8s_parity_test.go` | test asserting old `suCmd` shape | updated to use the helper |

## Helper design

Rather than open-coding the `sh -c if-then-else` block at each site, the fix introduces `runtime.ExecAsUserCmd(user, cmd string) []string`:

```go
func ExecAsUserCmd(user, cmd string) []string {
    const script = `if [ "$(whoami)" = "$1" ]; then exec sh -c "$2"; else exec su - "$1" -c "$2"; fi`
    return []string{"sh", "-c", script, "exec-as-user", user, cmd}
}
```

Three design notes:

1. **Both branches route the cmd through a fresh `sh -c`.** `exec` is a POSIX special builtin that does not parse leading `VAR=value` env-prefix assignments — without an inner shell, the PTY attach call sites that pass `TERM=xterm-256color tmux attach-session -t scion` would fail with exit code 127 (`exec: TERM=xterm-256color: not found`) once `whoami` matches the target user. Routing through `exec sh -c "$2"` makes both branches symmetric so any shell syntax valid in the `su -c` branch is also valid in the `whoami` branch.

2. **`user` and `cmd` are passed as positional shell arguments (`$1`/`$2`)** rather than interpolated into the script body via `fmt.Sprintf` + shell quoting. The shell's own argv machinery preserves them verbatim — no Go-side quoting layer that could disagree with what the inner shell expects. Cmds containing `$VAR`, backticks, or backslashes survive verbatim.

3. **`$0` is a script-name label (`"exec-as-user"`)** so anything that prints `$0` for diagnostics (shell error messages, `ps`) is self-describing.

## Defense-in-depth additions on the broker side

- A `sanitizeExecUser` helper in `pty_handlers.go` applies the same `^[a-zA-Z0-9_-]+$` regex that `KubernetesRuntime.Attach` uses for the username it derives from agent annotations. Invalid values fall back to `"scion"` with a logged warning. Used by both PTY constructors and `waitForTmuxSession`.
- The duplicated `"TERM=xterm-256color tmux attach-session -t scion"` literal in `pty_handlers.go` is extracted to a `tmuxAttachCmd` package const.

## Tests

- `ExecAsUserCmd` unit tests cover shape, single-quote preservation, username verbatim round-trip, an execution-based smoke test that proves the whoami branch runs the cmd directly, an env-prefix regression test that prevents reintroducing the `exec %s` bug, and a table-driven end-to-end test that runs each production cmd shape (env-prefixed, single-quoted token, plain, shell-quoted argv join) through the helper + `/bin/sh`.
- `sanitizeExecUser` table test covers empty fallback, valid users, shell metacharacter rejection, command substitution rejection, quoted-name rejection, and whitespace rejection.
- Existing tests in `k8s_podexec_test.go` and `k8s_parity_test.go` are updated to use the helper rather than hardcoding the shell string.

## Verification

End-to-end on GKE: the dashboard PTY now streams immediately once the tmux session is up, on agent images that previously hit the PAM-prompt bug.

Local CI:

- `make fmt-check` ✅
- `make lint` ✅
- `make build` ✅
- `make golangci-lint` ✅ (`0 issues` vs main)
- `go test ./pkg/runtime/... ./pkg/runtimebroker/...` ✅ (excluding the pre-existing macOS `/var` ↔ `/private/var` symlink failure in `TestCreateAgentGroveSlugInitializesScionDir`, which reproduces on bare `upstream/main`)

## Out of scope

- The unrelated `failed to look up existing grove members group by slug` warning that may appear in hub logs.
- The `Signing key not found in any source, generating new key` ephemeral-secrets bug.
- Any changes to agent container Dockerfiles, pod specs, or Workload Identity setup.
- Shell scripts in `image-build/` — strictly the Go call sites.

### DIFF
--- a/pkg/runtime/exec_user.go
+++ b/pkg/runtime/exec_user.go
@@ -1,0 +1,54 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package runtime
+
+// ExecAsUserCmd returns a /bin/sh command vector that runs cmd as user.
+//
+// If the container already runs as user (no su needed — common on GKE
+// Autopilot or any image with a non-root USER directive in the Dockerfile),
+// the wrapper execs cmd directly. Otherwise it falls back to
+// `su - user -c cmd` for legacy images that start as root.
+//
+// This avoids the PAM password prompt that occurs when root tries to su
+// to another user on container images whose /etc/pam.d/su lacks the
+// `auth sufficient pam_rootok.so` line — the default in node:20-slim and
+// other Debian-derived bases. On those images, a non-interactive
+// `su - <user> -c ...` exits with "Authentication failure" and the
+// caller (typically a readiness probe or PTY exec) misinterprets the
+// failure as the target service not being ready.
+//
+// See PR #159 for the original inline fix in KubernetesRuntime.Attach
+// that this helper generalizes to all su-wrapping call sites in the
+// broker and runtime packages.
+//
+// Both branches route the cmd through a fresh `sh -c` invocation, and
+// user/cmd are passed as positional shell arguments ($1/$2) rather
+// than interpolated into the script body. The shell's argv machinery
+// preserves them verbatim — no Go-side shell quoting, no risk of
+// double-quoted expansion in the outer wrapper. The inner shell
+// then parses cmd as a normal command line: leading env-var
+// assignments work (e.g. "TERM=xterm-256color tmux attach-session
+// -t scion"), shell metacharacters are interpreted by the inner
+// shell, and $VAR references resolve against that inner shell's
+// environment.
+func ExecAsUserCmd(user, cmd string) []string {
+	// Per `sh -c <script> <arg0> <arg1> ...`: arg0 becomes $0
+	// (conventionally the script name, which surfaces in ps and
+	// error messages), arg1 becomes $1, etc. We label $0 as
+	// "exec-as-user" so any diagnostic that prints it is
+	// self-describing.
+	const script = `if [ "$(whoami)" = "$1" ]; then exec sh -c "$2"; else exec su - "$1" -c "$2"; fi`
+	return []string{"sh", "-c", script, "exec-as-user", user, cmd}
+}
--- a/pkg/runtime/exec_user.go
+++ b/pkg/runtime/exec_user.go
@@ -14,6 +14,15 @@
 
 package runtime
 
+import "regexp"
+
+// ValidExecUserName matches usernames that are safe to interpolate
+// into shell command lines built by ExecAsUserCmd. Compiled once at
+// package init so callers (e.g. KubernetesRuntime.Attach) avoid a
+// per-invocation regexp.MustCompile, and so the broker's
+// sanitizeExecUser shares one source of truth for the rule.
+var ValidExecUserName = regexp.MustCompile(`^[a-zA-Z0-9_-]+$`)
+
 // ExecAsUserCmd returns a /bin/sh command vector that runs cmd as user.
 //
 // If the container already runs as user (no su needed — common on GKE
--- a/pkg/runtime/exec_user_test.go
+++ b/pkg/runtime/exec_user_test.go
@@ -236,3 +236,25 @@ func TestExecAsUserCmd_CallSiteShapes(t *testing.T) {
 		})
 	}
 }
+
+func TestValidExecUserName(t *testing.T) {
+	cases := []struct {
+		name string
+		in   string
+		want bool
+	}{
+		{name: "scion accepted", in: "scion", want: true},
+		{name: "alphanumeric with hyphen and underscore accepted", in: "agent-1_x", want: true},
+		{name: "empty rejected", in: "", want: false},
+		{name: "shell metachar rejected", in: "scion;rm -rf /", want: false},
+		{name: "command substitution rejected", in: "$(whoami)", want: false},
+		{name: "embedded space rejected", in: "two words", want: false},
+	}
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			if got := ValidExecUserName.MatchString(tc.in); got != tc.want {
+				t.Errorf("ValidExecUserName.MatchString(%q) = %v, want %v", tc.in, got, tc.want)
+			}
+		})
+	}
+}
--- a/pkg/runtime/exec_user_test.go
+++ b/pkg/runtime/exec_user_test.go
@@ -1,0 +1,238 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package runtime
+
+import (
+	"os/exec"
+	"os/user"
+	"runtime"
+	"strings"
+	"testing"
+)
+
+func TestExecAsUserCmd_Shape(t *testing.T) {
+	user := "scion"
+	cmd := "tmux has-session -t scion"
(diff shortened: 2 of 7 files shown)


ptone:
# PR #180 Review: fix(broker): apply #159's whoami-skip-su pattern to remaining su call sites

**Reviewer:** Automated Code Review (Senior Staff SWE / Security)
**Date:** 2026-04-25
**Branch:** `fix/whoami-skip-su-broker-paths` -> `main`
**Diff:** +428 / -39 across 7 files

---

## Executive Summary

This is a **low-risk, well-structured bug fix** that generalizes the PAM-skip-su pattern from PR #159 to all remaining `su -` call sites, fixing a real production failure (PTY stream never connects on GKE Autopilot). The security posture is improved: the new code eliminates per-invocation `regexp.MustCompile`, adds a defense-in-depth `sanitizeExecUser` guard on the broker side, and avoids Go-side shell interpolation via positional shell arguments.

---

## Critical Issues

**None.**

---

## Observations

### 1. Behavioral delta in `Attach` whoami branch: extra `sh -c` layer (Informational / No Action Required)

**Old code** (`k8s_runtime.go:1778`, base branch):
```go
`target=%q; if [ "$(whoami)" = "$target" ]; then exec tmux attach -t scion; else exec su - "$target" -c "tmux attach -t scion"; fi`
```
The whoami-matches branch ran `exec tmux attach -t scion` directly — zero intermediate shells.

**New code** (via `ExecAsUserCmd`):
```sh
if [ "$(whoami)" = "$1" ]; then exec sh -c "$2"; else exec su - "$1" -c "$2"; fi
```
The whoami-matches branch now runs `exec sh -c "$2"`, which briefly spawns an inner `sh` before `exec` replaces it. Since `exec` replaces the process immediately, the steady-state process tree is identical (only the `tmux` process survives). The extra shell is needed to make both branches symmetric for env-prefix commands (`TERM=xterm-256color ...`), which `exec` alone cannot parse. This is well documented in the PR description and the `ExecAsUserCmd` godoc. **No action needed.**

### 2. `ValidExecUserName` is an exported mutable global (`exec_user.go:24`)

```go
var ValidExecUserName = regexp.MustCompile(`^[a-zA-Z0-9_-]+$`)
```

This is a `var`, not a `const`-equivalent. Any package that imports `runtime` could reassign it, weakening the validation. While there's no evidence of misuse today and Go doesn't support `const` for `*regexp.Regexp`, the idiomatic pattern is a private var + public function:

```go
var validExecUserName = regexp.MustCompile(`^[a-zA-Z0-9_-]+$`)

func IsValidExecUserName(s string) bool {
    return validExecUserName.MatchString(s)
}
```

**Severity:** Low. The regex is already compiled correctly and both callers (the broker's `sanitizeExecUser` and `KubernetesRuntime.Attach`) use it read-only. This is a minor API-hardening improvement that doesn't block the fix.

### 3. `Exec` uses `r.ExecUser()` (always `"scion"`) — not annotation-derived username

In `k8s_runtime.go:2044`:
```go
suCmd := ExecAsUserCmd(r.ExecUser(), strings.Join(quoted, " "))
```

`ExecUser()` is hardcoded to `"scion"` on every runtime. Meanwhile, `Attach` reads `agent.Annotations["scion.username"]` and falls through to `"scion"` if absent. This asymmetry is **pre-existing** (the old `Exec` also hardcoded `"scion"`), so it's not introduced by this PR. However, if the annotation-derived username ever differs from `"scion"`, `Exec` would run commands as the wrong user. Worth a follow-up issue. **Not blocking.**

### 4. `sanitizeExecUser` logs unsanitized user value

In `pty_handlers.go:70`:
```go
slog.Warn("Invalid exec user, falling back to 'scion'", "user", user)
```

The raw invalid `user` string (which may contain shell metacharacters or injection attempts) is logged. This is safe — `slog` structured logging handles it correctly and the value is never evaluated by a shell — but it does mean untrusted input appears in logs. **No action needed**, but worth noting for compliance-sensitive deployments where log-injection is a concern.

### 5. Test: `TestExecAsUserCmd_Shape` check on line 63 has a subtle false-positive risk

```go
if strings.Contains(script, user) || strings.Contains(script, cmd) {
```

`user` is `"scion"` and `cmd` is `"tmux has-session -t scion"`. The check asserts neither appears in the script body. Since the script template is `if [ "$(whoami)" = "$1" ]; then ...`, this is correct today — but if someone renamed the label from `"exec-as-user"` to something containing `"scion"`, this test would break for the wrong reason. Very minor and unlikely to cause confusion in practice.

### 6. No remaining `su -` call sites (Verified)

A grep for `"su", "-"` across `pkg/` returns zero matches, confirming complete migration.

---

## Security Analysis

| Concern | Assessment |
|---|---|
| Shell injection via `user` | **Mitigated.** `ValidExecUserName` regex (`^[a-zA-Z0-9_-]+$`) rejects metacharacters. Broker additionally sanitizes via `sanitizeExecUser`. `ExecAsUserCmd` passes user as positional arg `$1`, avoiding interpolation. |
| Shell injection via `cmd` | **Not applicable at call sites.** All `cmd` values are either hardcoded string literals (`tmuxAttachCmd`, `"tmux has-session -t scion"`, etc.) or constructed from Go-side shell-quoting of `[]string` argv. No user-controlled input reaches `cmd`. |
| Regex DoS (ReDoS) | **Not a concern.** `^[a-zA-Z0-9_-]+$` is a simple character class with no alternation or nested quantifiers. |
| TOCTOU on `whoami` check | **Acceptable.** The `whoami` check and the `su`/`exec sh -c` are within the same shell script invocation. The UID cannot change between them without an external `nsenter`-level attack, which is outside the threat model. |

---

## Positive Feedback

1. **Excellent helper design.** Passing `user` and `cmd` as positional shell arguments (`$1`/`$2`) instead of `fmt.Sprintf` interpolation is the right pattern. It sidesteps an entire class of quoting bugs and is well-documented in the godoc.

2. **Symmetric branches.** Routing both the whoami and su branches through `sh -c` prevents the subtle `exec TERM=xterm-256color ...` failure that only the whoami branch would hit. The regression test (`TestExecAsUserCmd_EnvPrefixedCmd`) locks this in.

3. **`tmuxAttachCmd` const.** Extracting the duplicated string literal eliminates a maintenance hazard across the two `runK8sExec` methods.

4. **Defense-in-depth `sanitizeExecUser`.** The broker now validates the exec user at all three entry points (`newLocalPTYSession`, `NewStreamPTYHandler`, `waitForTmuxSession`), ensuring that values flowing into `ExecAsUserCmd` from agent metadata are always safe. The fail-open-to-"scion" behavior matches the pre-existing default.

5. **Thorough test coverage.** The test suite covers shape verification, shell-quoting preservation, username pass-through, env-prefix regression, table-driven call-site shapes, and runtime execution. The `sanitizeExecUser` tests cover all edge cases including command substitution and quoted names.

6. **Clean removal of the per-invocation `regexp.MustCompile`.** The old `Attach` code compiled the regex on every call. The new shared `ValidExecUserName` compiles once.

---

## Final Verdict

**Approve.** This PR fixes a real production bug, improves security posture, and is backed by comprehensive tests. The observations above are minor and none are blocking. The code is well-structured, idiomatic, and the PR description is exemplary in explaining the rationale and verification steps.

ptone:
Thank you for the contribution @jordantdavis

PR #206, opened 2026-04-30
fix(broker): resolve GCE hairpin NAT for colocated Docker bridge containers

ameer00:
## Summary

Fixes a networking issue where Docker bridge containers on a colocated hub+broker GCE VM cannot reach the hub's public domain due to GCE's hairpin NAT limitation. When a VM tries to reach its own external IP from within a Docker bridge container, the connection times out because GCE does not support hairpin NAT (traffic from a VM to its own external IP via the external interface).

This manifests as agents getting WebSocket 1006 disconnect errors immediately after starting, because the agent container cannot establish a connection back to the hub endpoint.

## Problem

In a **combo server** deployment (hub + runtime broker on the same GCE VM), the broker resolves the hub endpoint to a public domain (e.g. `hub.example.com` → `34.x.x.x`). When Docker bridge containers try to connect to this domain:

1. DNS resolves `hub.example.com` to the VM's own external IP (`34.x.x.x`)
2. The container sends traffic to `34.x.x.x` via the Docker bridge
3. GCE drops this traffic — it does not support hairpin NAT (a VM cannot reach its own external IP from within itself via the external interface)
4. The connection times out, causing agent WebSocket disconnections (1006 errors)

This does **not** affect:
- Kubernetes runtimes (pods use cluster networking)
- Non-colocated deployments (broker on a different machine than hub)
- Localhost endpoints (already handled by the existing `containerHubEndpoint` bridge override)

### Commit 1: `fix(broker): use localhost for co-located broker-to-hub communication`

When the hub and broker are co-located (same process in combo mode), the broker's internal HTTP calls to the hub were going through the public domain, which is unnecessary and triggers the hairpin NAT issue for the broker process itself. This commit ensures the broker uses `localhost` for its own hub communication when colocated, via `resolveHubConnection()`.

**Files changed:**
- `cmd/server_foreground.go` — Refactored `IsColocated` detection to use `resolveHubConnection()` and pass the connection context through to the broker
- `cmd/server_bridge_test.go` — Added `TestResolveHubConnection` with test cases covering colocated detection from request headers

### Commit 2: `fix(broker): add --add-host for hub domain in colocated Docker bridge containers`

This is the main fix. When the broker detects it is colocated with the hub **and** the runtime is Docker/Podman, it adds `--add-host <hub-domain>:host-gateway` to the container's run configuration. This maps the hub's public domain to `host-gateway` (Docker's alias for the host's bridge IP, typically `172.17.0.1`), allowing the container to reach the hub via the Docker bridge instead of trying to go through the external IP.

**Files changed:**
- `pkg/runtimebroker/hubenv.go` — Added `colocatedExtraHosts()` function that computes `--add-host` entries. Skips: non-colocated, Kubernetes, localhost endpoints, raw IP addresses, empty endpoints
- `pkg/runtimebroker/start_context.go` — Moved `resolveHubConnection()` earlier in `buildStartContext()` (before hub endpoint resolution) so `IsColocated` is available. Added colocated detection and `extraHosts` computation, assigned to `opts.ExtraHosts`
- `pkg/api/types.go` — Added `ExtraHosts []string` field to `StartOptions` to carry broker-computed host entries through to the runtime layer
- `pkg/agent/run.go` — Changed `ExtraHosts` assignment to merge broker-provided hosts with existing bridge extra hosts via new `mergeExtraHosts()` helper. Broker entries take precedence on hostname conflicts
- `pkg/runtimebroker/hubenv_test.go` — Added `TestColocatedExtraHosts` (6 test cases) and `TestRedactEnvValueForLog`
- `pkg/agent/run_metadata_test.go` — Added `TestMergeExtraHosts` (5 test cases covering nil inputs, no overlap, and precedence on overlap)

## How it works

```
Before (broken):
  Container → DNS(hub.example.com) → 34.x.x.x (external IP) → DROPPED by GCE

After (fixed):
  Container → /etc/hosts(hub.example.com → 172.17.0.1) → Docker bridge → host → localhost:8080 → hub
```

The `--add-host hub.example.com:host-gateway` entry in the container's `/etc/hosts` overrides DNS resolution, routing traffic through the Docker bridge to the host instead of through the external IP.

## Verified on

- **Test hub**: `scion-test01` (GCE VM, combo mode, Docker runtime) at `hub.test01.scionduet03.ameer.cloud`
  - Container inspect confirms `ExtraHosts: ["hub.test01.scionduet03.ameer.cloud:host-gateway"]`
  - Container `/etc/hosts` shows `172.17.0.1 hub.test01.scionduet03.ameer.cloud`
  - `curl https://hub.test01.scionduet03.ameer.cloud/healthz` from inside container returns healthy
  - Agents start successfully, maintain WebSocket connections, and send heartbeats

## Test plan

- [x] `TestColocatedExtraHosts` — 6 cases: public domain, localhost (skipped), kubernetes (skipped), not colocated (skipped), IP address (skipped), empty endpoint (skipped)
- [x] `TestMergeExtraHosts` — 5 cases: nil inputs, single input, no overlap, overlap with precedence
- [x] `TestResolveHubConnection` — colocated detection from request headers
- [x] Existing `TestApplyContainerBridgeOverride`, `TestResolveHubEndpointForStartPrecedence`, `TestResolveHubEndpointForCreatePrecedence` all pass
- [x] End-to-end: deployed to test GCE VM, created agents, verified container networking

### DIFF
--- a/cmd/server_bridge_test.go
+++ b/cmd/server_bridge_test.go
@@ -171,6 +171,93 @@ func TestResolveHubEndpointSettingsFallback(t *testing.T) {
 	})
 }
 
+func TestResolveHubEndpointForBroker(t *testing.T) {
+	origEnableHub := enableHub
+	origEnableDebug := enableDebug
+	origEnableWeb := enableWeb
+	origWebPort := webPort
+	defer func() {
+		enableHub = origEnableHub
+		enableDebug = origEnableDebug
+		enableWeb = origEnableWeb
+		webPort = origWebPort
+	}()
+
+	t.Run("co-located always uses localhost even with SCION_SERVER_BASE_URL", func(t *testing.T) {
+		enableHub = true
+		enableDebug = false
+		enableWeb = true
+		webPort = 8080
+		t.Setenv("SCION_SERVER_BASE_URL", "https://scionduet03.ameer.cloud")
+
+		cfg := &config.GlobalConfig{}
+		settings := &config.Settings{}
+
+		got := resolveHubEndpointForBroker(cfg, settings)
+		assert.Equal(t, "http://localhost:8080", got)
+	})
+
+	t.Run("co-located uses hub port when web disabled", func(t *testing.T) {
+		enableHub = true
+		enableDebug = false
+		enableWeb = false
+		webPort = 8080
+
+		cfg := &config.GlobalConfig{
+			Hub: config.HubServerConfig{Port: 9810},
+		}
+		settings := &config.Settings{}
+
+		got := resolveHubEndpointForBroker(cfg, settings)
+		assert.Equal(t, "http://localhost:9810", got)
+	})
+
+	t.Run("explicit RuntimeBroker.HubEndpoint takes priority", func(t *testing.T) {
+		enableHub = true
+		enableDebug = false
+		enableWeb = true
+		webPort = 8080
+
+		cfg := &config.GlobalConfig{
+			RuntimeBroker: config.RuntimeBrokerConfig{
+				HubEndpoint: "https://custom-hub.example.com",
+			},
+		}
+		settings := &config.Settings{}
+
+		got := resolveHubEndpointForBroker(cfg, settings)
+		assert.Equal(t, "https://custom-hub.example.com", got)
+	})
+
+	t.Run("non-hub mode uses settings endpoint", func(t *testing.T) {
+		enableHub = false
+		enableDebug = false
+		enableWeb = false
+
+		cfg := &config.GlobalConfig{}
+		settings := &config.Settings{
+			Hub: &config.HubClientConfig{
+				Endpoint: "https://remote-hub.example.com",
+			},
+		}
+
+		got := resolveHubEndpointForBroker(cfg, settings)
+		assert.Equal(t, "https://remote-hub.example.com", got)
+	})
+
+	t.Run("non-hub mode returns empty when no settings", func(t *testing.T) {
+		enableHub = false
+		enableDebug = false
+		enableWeb = false
+
+		cfg := &config.GlobalConfig{}
+		settings := &config.Settings{}
+
+		got := resolveHubEndpointForBroker(cfg, settings)
+		assert.Equal(t, "", got)
+	})
+}
+
 func TestIsLocalhostURL(t *testing.T) {
 	tests := []struct {
 		endpoint string
--- a/cmd/server_foreground.go
+++ b/cmd/server_foreground.go
@@ -1005,7 +1005,7 @@ func startRuntimeBroker(ctx context.Context, cmd *cobra.Command, cfg *config.Glo
 	slog.SetDefault(slog.Default().With(slog.String(logging.AttrBrokerID, brokerID)))
 
 	// Resolve hub endpoint for the runtime broker
-	hubEndpointForRH := resolveHubEndpointForBroker(cfg, hubEndpoint, settings)
+	hubEndpointForRH := resolveHubEndpointForBroker(cfg, settings)
 
 	// Auto-provide defaults
 	if enableHub && !cmd.Flags().Changed("auto-provide") {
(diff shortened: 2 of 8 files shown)


ptone:
# PR #206 Review: fix(broker): resolve GCE hairpin NAT for colocated Docker bridge containers

**Author:** ameer00  
**Branch:** `fix/colocated-docker-bridge-hairpin-nat` → `main`  
**Files changed:** 8 (+265 / -25)  
**Reviewed:** 2026-05-01

---

## Executive Summary

This PR fixes a real networking issue where Docker bridge containers on a colocated hub+broker GCE VM cannot reach the hub's public domain due to GCE's lack of hairpin NAT support. The fix is well-scoped, defensively coded, and thoroughly tested — it introduces no new security concerns and carries low regression risk.

---

## Critical Issues

**None found.**

---

## Observations

### 1. `mergeExtraHosts` hostname extraction with colons (Minor — Low Risk)

**File:** `pkg/agent/run.go:1214`

```go
host, _, _ := strings.Cut(h, ":")
```

`strings.Cut` splits on the *first* colon. For the expected input format `hostname:target` (e.g., `hub.example.com:host-gateway`), this is correct. However, if an entry ever contained an IPv6 literal like `[::1]:host-gateway`, the extracted "host" would be `[`, not `[::1]`, leading to incorrect deduplication.

**Risk:** Very low. Upstream callers (`colocatedExtraHosts` and `BridgeExtraHosts`) both produce DNS-name-based entries, never raw IPv6 literals. Additionally, `colocatedExtraHosts` explicitly filters out IP addresses via `net.ParseIP(host) != nil`. This is a theoretical concern, not a practical one.

**Suggested defensive hardening (optional):**
```go
// If needed in the future, use net.SplitHostPort for entries
// with brackets, but current callers guarantee DNS-name format.
```

### 2. Podman runtime not excluded from `colocatedExtraHosts` (Correct Behavior)

**File:** `pkg/runtimebroker/hubenv.go:131`

The function only skips `kubernetes` but allows `docker`, `podman`, and `apple_container` runtimes through. This is correct — Podman (4.1+) supports `--add-host hostname:host-gateway` identically to Docker, and the hairpin NAT issue applies equally to both. No change needed.

### 3. `hubConn` resolution moved earlier — ordering correctness verified

**File:** `pkg/runtimebroker/start_context.go:221-226`

The `resolveHubConnection()` call was moved from after template hydration (line ~356 in the base) to before hub endpoint resolution (line ~221). This is safe because:
- `resolveHubConnection` only reads from `s.hubConnections` (immutable by this point in the request lifecycle) and the request header `X-Scion-Hub-Connection`.
- It has no dependency on `hubEndpoint`, `env`, or any state computed later in `buildStartContext`.
- The original later call site is now removed, avoiding a redundant map lookup.

### 4. Start/restart path always has `HTTPRequest` set

All four call sites of `buildStartContext` in `handlers.go` pass `HTTPRequest: r`, so the `in.HTTPRequest != nil` guard on line 224 will always be true in production. The nil-safety is good defensive coding for future callers or test usage.

### 5. Removal of `hubEndpoint` parameter from `resolveHubEndpointForBroker` — verified safe

**File:** `cmd/server_foreground.go:1261`

The old code had a path where the broker's hub endpoint could inherit from a previously resolved `hubEndpoint` variable (potentially set from `SCION_SERVER_BASE_URL` — a public URL). The new code always constructs `http://localhost:<port>` in colocated mode. This is both the root-cause fix for the broker's own connections and is clearly the correct behavior: a colocated broker should never route through the external URL for internal IPC.

The explicit `cfg.RuntimeBroker.HubEndpoint` override still takes priority (line 1262), preserving intentional user configuration.

### 6. Test coverage is thorough

- `TestColocatedExtraHosts` — 6 cases covering all skip conditions (localhost, kubernetes, not-colocated, IP address, empty endpoint) plus the positive case.
- `TestMergeExtraHosts` — 5 cases including nil inputs, no overlap, and precedence on overlap.
- `TestResolveHubEndpointForBroker` — 5 cases verifying localhost-always behavior in colocated mode, explicit override priority, and non-hub-mode fallback.
- All tests pass cleanly.

---

## Positive Feedback

1. **Excellent PR description.** The problem, fix, and verification are all clearly documented with a before/after network flow diagram. This is a model PR description.

2. **Layered fix approach.** Commit 1 fixes the broker's own communication (always use localhost when colocated), and Commit 2 fixes container communication (add `--add-host` for the hub domain). These are logically distinct and independently valuable.

3. **Defensive guard clauses in `colocatedExtraHosts`.** The function has an explicit chain of early returns for every case where extra hosts should *not* be added (not colocated, kubernetes, localhost, IP addresses, empty endpoint, parse errors). This fail-closed approach is exactly right for network-affecting configuration.

4. **`mergeExtraHosts` precedence design.** Broker-provided entries (`a`) take precedence over bridge entries (`b`) on hostname conflict, which is the correct semantic — the colocated fix should override the generic bridge entry, not the other way around.

5. **No semantic change to the start path for non-colocated deployments.** The extra hosts computation is a no-op when `isColocated` is false, so Kubernetes and remote-broker deployments are completely unaffected.

---

## Final Verdict

**Approve.** This is a clean, well-tested fix for a real infrastructure issue. The changes are minimal, correctly scoped to the affected code path, and the defensive coding style (fail-closed guards, explicit precedence) inspires confidence. No blocking issues found.

PR #208, opened 2026-05-04
add directory to opencode system-prompt/instructions writing

verdverm:
Makes the file location consistent with other harnesses, was not being picked up by `opencode` because it ended up in `$HOME` instead of `$HOME/.config/opencode`


ptone:
opencode system-prompt, target file deleted by #336

PR #214, opened 2026-05-05
opencode: support vertex-ai auth

verdverm:
1. Adds support to `opencode` for vertex-ai auth, tested "in production"
2. fixes logic around `--prompt` flag when no task consuming next arg from baseArgs as prompt, i.e. if no task, should not even supply the flag

### DIFF
--- a/pkg/harness/opencode.go
+++ b/pkg/harness/opencode.go
@@ -50,7 +50,7 @@ func (o *OpenCode) AdvancedCapabilities() api.HarnessAdvancedCapabilities {
 		Auth: api.HarnessAuthCapabilities{
 			APIKey:   api.CapabilityField{Support: api.SupportYes},
 			AuthFile: api.CapabilityField{Support: api.SupportYes},
-			VertexAI: api.CapabilityField{Support: api.SupportNo, Reason: "Vertex AI auth is not supported for this harness"},
+			VertexAI: api.CapabilityField{Support: api.SupportYes},
 		},
 		Resume: api.CapabilityField{Support: api.SupportYes},
 	}
@@ -64,11 +64,8 @@ func (o *OpenCode) GetCommand(task string, resume bool, baseArgs []string) []str
 	args := []string{"opencode"}
 	if resume {
 		args = append(args, "--continue")
-	} else {
-		args = append(args, "--prompt")
-		if task != "" {
-			args = append(args, task)
-		}
+	} else if task != "" {
+		args = append(args, "--prompt", task)
 	}
 
 	args = append(args, baseArgs...)
@@ -142,12 +139,22 @@ func (o *OpenCode) ResolveAuth(auth api.AuthConfig) (*api.ResolvedAuth, error) {
 					{SourcePath: auth.OpenCodeAuthFile, ContainerPath: "~/.local/share/opencode/auth.json"},
 				},
 			}, nil
+		case "vertex-ai":
+			if auth.GoogleCloudProject == "" || auth.GoogleCloudRegion == "" {
+				return nil, fmt.Errorf("opencode: auth type %q selected but GOOGLE_CLOUD_PROJECT and/or GOOGLE_CLOUD_REGION not set", auth.SelectedType)
+			}
+			return o.resolveVertexAI(auth), nil
+
 		default:
-			return nil, fmt.Errorf("opencode: unknown auth type %q; valid types are: api-key, auth-file", auth.SelectedType)
+			return nil, fmt.Errorf("opencode: unknown auth type %q; valid types are: api-key, auth-file, vertex-ai", auth.SelectedType)
 		}
 	}
 
-	// Auto-detect preference order: AnthropicAPIKey → OpenAIAPIKey → OpenCodeAuthFile → error
+	// Auto-detect preference order: VertexAi → AnthropicAPIKey → OpenAIAPIKey → OpenCodeAuthFile → error
+
+	if auth.GoogleCloudProject != "" && auth.GoogleCloudRegion != "" {
+		return o.resolveVertexAI(auth), nil
+	}
 
 	if auth.AnthropicAPIKey != "" {
 		return &api.ResolvedAuth{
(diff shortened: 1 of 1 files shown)


ptone:
# PR Review: opencode: support vertex-ai auth

**PR:** #214
**Branch:** (PR head) → `main`
**Commit:** `d0dfe9e6`
**Author:** verdverm
**Reviewer:** Scion Agent (pr-214)
**Date:** 2026-06-01

---

## Executive Summary

**Medium risk — request changes.** Enables Vertex AI as an auth method for the OpenCode harness by flipping the capability flag, adding a `vertex-ai` branch in `ResolveAuth`, introducing auto-detection on `GOOGLE_CLOUD_PROJECT` + `GOOGLE_CLOUD_REGION`, and adding a `resolveVertexAI` helper that mirrors `claude_code.go:430`. The shape is right, but the implementation diverges from the Claude Code precedent in ways that will bite GCP users: Vertex is auto-detected *ahead* of API keys, and detection doesn't require any actual credentials. Also bundles an unrelated `GetCommand` cleanup and ships with zero tests for the new path.

---

## Files Changed (1 file, +34 / -9)

| File | Change |
|------|--------|
| `pkg/harness/opencode.go` | Capability flip, `vertex-ai` selection + auto-detect, new `resolveVertexAI`, bundled `GetCommand` refactor |

---

## Critical / Blocking Issues

### 1. Auto-detect priority puts Vertex AI ahead of API keys — silent behavior regression

**File:** `pkg/harness/opencode.go:152`

```go
// Auto-detect preference order: VertexAi → AnthropicAPIKey → OpenAIAPIKey → OpenCodeAuthFile → error

if auth.GoogleCloudProject != "" && auth.GoogleCloudRegion != "" {
    return o.resolveVertexAI(auth), nil
}

if auth.AnthropicAPIKey != "" { ... }
```

Compare with the Claude Code harness, which intentionally puts Vertex **last** in the auto-detect chain:

**`claude_code.go:384`:** `// Auto-detect preference order: API key → OAuth token → credentials file → Vertex AI → error`

Many GCP users have `GOOGLE_CLOUD_PROJECT` / `GOOGLE_CLOUD_REGION` exported in their shell environment for unrelated reasons (gcloud CLI, other tooling) while also setting `ANTHROPIC_API_KEY`. After this PR they'll silently flip to Vertex on the next opencode run. The PR also doesn't update the existing `TestOpenCodeResolveAuth_PreferenceOrder` test, which would have caught this behavior change.

**Suggested fix:** Move the Vertex AI block to *after* the auth-file block, matching the Claude Code precedence. Update the comment accordingly.

### 2. Auto-detect doesn't gate on actual credentials being present

**File:** `pkg/harness/opencode.go:152-154`

```go
if auth.GoogleCloudProject != "" && auth.GoogleCloudRegion != "" {
    return o.resolveVertexAI(auth), nil
}
```

`resolveVertexAI` will happily return success with an empty `Files` slice if `GoogleAppCredentials` is unset. The container starts, opencode tries to authenticate against Vertex, and fails at runtime with a less actionable error than "no valid auth method found."

Compare with `claude_code.go:422-424`:

```go
hasVertexCreds := auth.GoogleAppCredentials != "" || auth.GCPMetadataMode == "assign"
if hasVertexCreds && auth.GoogleCloudProject != "" && auth.GoogleCloudRegion != "" {
    return c.resolveVertexAI(auth), nil
}
```

**Suggested fix:** Add the same `hasVertexCreds` precondition. Apply analogous validation in the explicit `case "vertex-ai":` branch (`opencode.go:142`) — currently it accepts a selection with no credentials mounted.

### 3. Env vars likely won't activate Vertex in opencode

**File:** `pkg/harness/opencode.go:200-209`

```go
EnvVars: map[string]string{
    "VERTEX_LOCATION":      auth.GoogleCloudRegion,
    "GOOGLE_CLOUD_REGION":  auth.GoogleCloudRegion,
    "GOOGLE_CLOUD_PROJECT": auth.GoogleCloudProject,
},
```

Two concerns:

**(a)** opencode is built on the Vercel AI SDK. Its `@ai-sdk/google-vertex` provider reads `GOOGLE_VERTEX_LOCATION` / `GOOGLE_VERTEX_PROJECT` (or is configured via `opencode.json` provider block), not `VERTEX_LOCATION`. The embedded config at `pkg/harness/opencode/embeds/opencode.json` is:

```json
{
  "$schema": "https://opencode.ai/config.json",
  "theme": "matrix"
}
```

No provider block, no model selection. Without `opencode.json` config or the SDK-expected env names, the env vars set here may not be picked up at all.

**(b)** `gemini_cli.go:376-378` sets *both* `GOOGLE_CLOUD_REGION` and `GOOGLE_CLOUD_LOCATION` because Google tooling disagrees on the canonical name (`GOOGLE_CLOUD_LOCATION` is the GenAI SDK convention). This PR only sets `GOOGLE_CLOUD_REGION`.

**Suggested fix:** Smoke-test end-to-end with a real opencode + Vertex run. Verify which env vars opencode actually reads and/or add a provider block to the embedded `opencode.json`. Mirror gemini_cli's belt-and-suspenders region/location handling.

---

## Observations

### 4. No test coverage for the new path

`opencode_test.go` has dedicated tests for `AnthropicAPIKey`, `OpenAIAPIKey`, `AuthFile`, preference order, and the no-creds error, but **nothing for vertex-ai**. The existing `TestOpenCodeResolveAuth_NoCreds` only asserts the error mentions `ANTHROPIC_API_KEY`, so it still passes even though the error message changed.

Tests to add:
- Explicit `SelectedType: "vertex-ai"` happy path.
- Explicit `vertex-ai` with missing project or region (expects error).
- Auto-detect via `GoogleCloudProject` + `GoogleCloudRegion` (and credentials — see #2).
- `GoogleAppCredentials` produces a `FileMapping` with the right container path.
- Preference ordering (Vertex vs API key — wire this to whichever order is decided in #1).

### 5. Bundled, unrelated `GetCommand` refactor

**File:** `pkg/harness/opencode.go:63-72`

The `if/else` collapse is a separate behavior change: previously `--prompt` was appended even when `task == ""`, which would silently consume the first `baseArgs` entry as the prompt — a latent bug. The new code skips `--prompt` entirely in that case, which is probably correct. But it has nothing to do with Vertex AI auth and deserves its own commit and a regression test (`TestOpenCodeGetCommand_EmptyTask` or similar).

---

## Nits

- **`opencode.go:208-209`:** Missing blank line between `resolveVertexAI` and `InjectSystemPrompt`.
- **`opencode.go:150`:** Comment says `VertexAi` — should be `VertexAI` for consistency with the codebase.
- **`opencode.go:189`:** New error message `"set VertexAi ENVs or ANTHROPIC_API_KEY..."` is vague. `claude_code.go:427` is the model — name the actual env vars (`GOOGLE_CLOUD_PROJECT`, `GOOGLE_CLOUD_REGION`, ADC path).

---

## Security

No new secret-handling surface. ADC file is mounted to the standard container path (`~/.config/gcloud/application_default_credentials.json`), same as `claude_code.go`. No issues.

---

## Final Verdict

**Request changes.** The capability + helper structure is correct, but three issues should block merge:

1. **Reorder** auto-detect so Vertex is last (parity with `claude_code.go`) to avoid silent regression for users with GCP env vars and an API key.
2. **Gate** auto-detect (and ideally explicit selection) on actual credentials being present, not just project + region.
3. **Verify** the env vars actually activate Vertex in opencode — embedded `opencode.json` may need a provider block, and the env var names should be confirmed against the AI SDK.

Tests for the new path should land in the same PR. The `GetCommand` cleanup is a nice fix but belongs in its own commit.

verdverm:
@ptone This was an early pass on this and noticed auto-detection picks the first auth it finds, which is why I put the vertex first. I have since added a parameter/flag to disable host-level credential passthrough and only use configured auth. I was unable to get vertex to work with anthropic creds on my host if we only change the order above.

Thoughts? (did I see some PR/changeset to disable host auth passthrough go by?)

In agreement with the other PR feedback.

I can pull out the code that disables host-auth passthrough and open a separate PR. Should that come first?

ptone:
sorry for slow reply. given opencode has auth libraries this should be just a matter of detecting the cloud settings are available

vertex-ai should autodetect if:

(GOOGLE_CLOUD_REGION OR GOOGLE_CLOUD_LOCATION OR
VERTEX_LOCATION
) is available

&& 

(GOOGLE_CLOUD_PROJECT OR
VERTEXAI_PROJECT) is available

&&

gcp service account != BLOCK

final keys in env should be:
VERTEXAI_PROJECT
VERTEX_LOCATION


that quasi spec may be all a coding agent needs.

ptone:
i used the above quasi spec to generate

https://github.com/GoogleCloudPlatform/scion/compare/main...ptone:scion/opencode-vertex-auth

but have not really run it through any test yet. 

i've also been working on some UX to launch harness with "no-auth" and then capture the auth details in initial interactive mode and then use those captured credentials for subsequent agents of that type.

ptone:
Functionality merged in #493 thank you for the contribution and helping steer what gets shipped

PR #216, opened 2026-05-06
Scion-a2a bridge

ptone:
Adds an A2A bridge to scion


ptone:
# PR #216 Review — Scion-A2A Bridge

- **PR**: https://github.com/GoogleCloudPlatform/scion/pull/216
- **Author**: @ptone
- **Branch**: `scion/a2a-bridge` → `main`
- **Size**: +6,703 / −0 across 25 new files (new module under `extras/scion-a2a-bridge/`)

## Executive Summary

This PR introduces a brand-new, self-contained module that bridges the A2A (Agent-to-Agent) protocol to the Scion Hub. The scaffolding (HTTP server, JSON-RPC dispatch, SQLite state, broker plugin wiring, metrics, rate limiting, push/SSE) is well-structured and covered by reasonable unit tests; however, several **correctness and security defects in the core dispatch, push, and auth paths are blocking** for a production deployment. The module is isolated under `extras/`, so the blast radius for the rest of the repo is small, but the bridge itself should not be merged in its current shape.

---

## Critical Issues

### C1. Cross-talk between concurrent SendMessage callers (data leak / wrong-recipient delivery)

**File**: `internal/bridge/bridge.go:75-180`, `253-347`

`Bridge.waiters` and `Bridge.activeTasks` are keyed only by `agentSlug`, not by task ID:

```go
b.addWaiter(agentCtx.AgentSlug, responseCh)        // SendMessage
…
waiters := b.waiters[agentSlug]                    // HandleBrokerMessage
for _, ch := range waiters {
    select {
    case ch <- msg:
    default:
    }
}
```

If two A2A clients call `message/send` against the same agent concurrently:

1. Both register a waiter under the same `agentSlug` key.
2. The first response from the agent is broadcast to **both** waiters (or non-deterministically to one — `default:` drops if the buffered channel is full, but with capacity 1 both can succeed).
3. There is no correlation back to the originating task / message ID, so the wrong client may receive the wrong agent's reply, and the second client will likely time out waiting for a response that was already consumed.

The same flaw applies to `activeTasks[agentSlug] []string` in `HandleBrokerMessage` — every active task for that agent gets the same broker message attributed to it, so multiple SSE/push subscribers see each other's results.

**Suggested fix** — propagate a correlation identifier end-to-end. The Scion `messages.StructuredMessage` should carry the originating A2A `taskId` (e.g., in metadata) and `HandleBrokerMessage` must dispatch only to the matching waiter / active task.

```go
// On send:
scionMsg.Metadata = map[string]string{"a2aTaskId": taskID}

// In HandleBrokerMessage:
if id := msg.Metadata["a2aTaskId"]; id != "" {
    b.dispatchToTask(id, msg)
    return nil
}
```

If the underlying broker has no per-task topic, at minimum key on `(agentSlug, taskID)` and tag outbound messages with the taskID so inbound responses can be routed back. Without this, the bridge cannot safely serve more than one concurrent caller per agent.

---

### C2. Push-notification SSRF — DNS rebinding bypass

**File**: `internal/bridge/push.go:48-100, 147-192`

`ValidatePushURL` and `send()` resolve the host with `net.LookupIP` and reject private IPs, but the actual HTTP request is made via `pd.client.Do(req)`, which **re-resolves the hostname inside `net/http.Transport`**. The IPs validated in `send()` are never pinned, so an attacker controlling DNS for a hostname can:

1. Have the validation lookup return a public IP (`203.0.113.1`).
2. Have the `Transport` lookup return `169.254.169.254` (or `127.0.0.1`, or any RFC1918 address).

This defeats the SSRF guard and is the canonical DNS-rebinding attack. The redirect blocker (`errRedirectBlocked`) does not help here.

**Suggested fix** — install a custom `Dialer.Control` (or `DialContext`) on the HTTP client that inspects the resolved address at connection time and rejects private/reserved ranges. Example:

```go
dialer := &net.Dialer{
    Control: func(network, address string, c syscall.RawConn) error {
        host, _, _ := net.SplitHostPort(address)
        ip := net.ParseIP(host)
        if ip == nil || isPrivateIP(ip) {
            return ErrSSRFBlocked
        }
        return nil
    },
}
client := &http.Client{
    Timeout: 10 * time.Second,
    Transport: &http.Transport{DialContext: dialer.DialContext},
    CheckRedirect: ...,
}
```

This pins the SSRF check to the actual TCP connection, eliminating the TOCTOU window.

---

### C3. SSRF allow-list is IPv4-only and misses ULA / metadata-v6

**File**: `internal/bridge/push.go:67-85`

The CIDR list in `isPrivateIP` only contains IPv4 ranges. While `IsLoopback`/`IsLinkLocalUnicast` catches `::1` and `fe80::/10`, the following are **not** blocked:

- IPv6 ULA: `fc00::/7`
- IPv6 site-local (deprecated but seen): `fec0::/10`
- IPv4 multicast/broadcast: `224.0.0.0/4`, `255.255.255.255/32`
- IPv4 CGNAT: `100.64.0.0/10`
- IPv6 EC2 IMDS: `fd00:ec2::254`

Combine with C2 and the SSRF guard becomes largely cosmetic.

**Suggested fix** — extend the block list to include IPv6 ULA and CGNAT, and use `ip.IsPrivate()` (Go 1.17+) as the primary check, augmented by explicit reserved ranges.

---

### C4. Auth fails open when API key is unset

**File**: `internal/bridge/server.go:529-558`

```go
if s.config.Auth.APIKey == "" {
    next.ServeHTTP(w, r)
    return
}
```

If the operator forgets to populate `auth.api_key` (or `${A2A_API_KEY}` env-var expansion silently produces empty string — see C7), **the entire JSON-RPC surface, including push-notification registration and message send, becomes unauthenticated**. There is no startup-time validation. This is the classic "fail open" misconfiguration.

**Suggested fix** — fail closed. In `loadConfig` (or `Server` construction), if `cfg.Auth.Scheme != ""` and a static key is required but empty, refuse to start:

```go
if cfg.Auth.Scheme == "apiKey" && cfg.Auth.APIKey == "" {
    return nil, fmt.Errorf("auth.api_key required when scheme is apiKey")
}
```

Add a separate explicit `auth.disabled: true` knob if running un-authenticated is ever desired.

---

### C5. WriteTimeout kills SSE streams after 120 s

**File**: `cmd/scion-a2a-bridge/main.go:136-142`

```go
httpServer := &http.Server{
    …
    WriteTimeout: 120 * time.Second,
}
```

`http.Server.WriteTimeout` applies to the **whole response**, not per-write. SSE streams (`message/stream`, `tasks/resubscribe`) are explicitly long-lived and will be terminated 120 s after first byte regardless of keepalives. The advertised `streaming: true` capability will misbehave for any task longer than 2 minutes.

**Suggested fix** — set `WriteTimeout: 0` on this server (relying on context/cancellation and SSE keepalives for liveness), or split SSE routes onto a dedicated `http.Server` with `WriteTimeout: 0` and `IdleTimeout` instead.

---

### C6. Push subscription is **deleted** after one batch of failures — silent unsubscribe

**File**: `internal/bridge/push.go:121-145`

```go
pd.log.Error("push notification exhausted retries, removing config", …)
pd.store.DeletePushConfig(cfg.ID)
```

A transient outage on the receiver (or a brief network blip) will:

1. Burn through `PushRetryMax` retries (default 3, with exponential backoff topping out around 8 s).
2. Permanently delete the subscription from SQLite.
3. Leave the A2A client with no notification that their webhook was de-registered.

This is hostile to clients and will manifest as random "the bridge stopped calling me" failures.

**Suggested fix** — only delete on a **persistent** `4xx` (`410 Gone`, repeated `400`), not on `5xx` or network errors. For other failures, leave the row in place and rely on the next dispatch attempt or operator intervention. Consider tracking `consecutive_failures` and only disabling after a much higher threshold.

---

### C7. `os.ExpandEnv` over the entire YAML corrupts legitimate `$` content

**File**: `cmd/scion-a2a-bridge/main.go:184`

```go
expanded := os.ExpandEnv(string(data))
```

This replaces every `$VAR` and `${VAR}` in the file. Any legitimate `$` in a password, prompt, or URL fragment becomes a silent empty string when no matching env var exists. Worse: `${A2A_API_KEY}` evaluating to empty silently disables auth (see C4).

**Suggested fix** — use `os.Expand` with a mapping function that errors on missing variables, and either restrict expansion to specific fields or document the constraint loudly:

```go
expanded := os.Expand(string(data), func(name string) string {
    v, ok := os.LookupEnv(name)
    if !ok {
        missing = append(missing, name)
    }
    return v
})
if len(missing) > 0 {
    return nil, fmt.Errorf("config references unset env vars: %v", missing)
}
```

---

### C8. CancelTask never interrupts the agent

**File**: `internal/bridge/bridge.go:219-251`

`CancelTask` updates the local SQLite state to `canceled` and emits a final SSE event, but never tells the Scion Hub to stop the agent. The `hubclient.Agents().SendStructuredMessage` call already takes an `interrupt bool` parameter (verified at `pkg/hubclient/agents.go:432`), but it is not used here. The agent will continue executing — and its subsequent reply will be **delivered to the next caller** of the same agent (see C1), now wrongly attributed.

**Suggested fix** — call `SendStructuredMessage(ctx, agentID, interruptMsg, true /*interrupt*/, false)` (or whatever the Hub-side cancel API is) before marking the task canceled.

---

## Observations

### O1. `GenerateAgentCard` and `resolveContext` list **all** Hub agents on every call

**File**: `internal/bridge/bridge.go:423-443, 470-484`

Both code paths call `b.hubClient.Agents().List(ctx, nil)` and linear-scan the result. The agent-card endpoint is **public** (`authMiddleware` skips `/.well-known/agent-card.json`), so an unauthenticated attacker can amplify load on the Hub at line rate. With many agents, every card request is also O(N) on the bridge.

Recommend a `GetBySlug` Hub call (or a small TTL cache keyed by `(grove, slug)`) and cache enrichment for at least the duration of the `Cache-Control: public, max-age=300` you already advertise.

### O2. Unbounded goroutine fan-out per push event

**File**: `internal/bridge/push.go:104-119`

```go
for _, cfg := range configs {
    go pd.sendWithRetry(cfg, event)
}
```

A task with many push subscribers and a bursty agent will spawn an unbounded goroutine count, each holding a 10 s HTTP timeout plus retry backoff. Use a worker pool or `errgroup` with a semaphore.

### O3. `streams.Broadcast` silently drops events when subscriber channel is full

**File**: `internal/bridge/stream.go:101-113`

```go
select {
case ch <- event:
default:
}
```

Channel buffer is 16. Under load (status-change storm during long execution), events are dropped without record. SSE clients will see truncated state transitions. At minimum, log dropped events and consider increasing the buffer or using a per-subscriber drop counter exposed via metrics.

### O4. Background goroutines from `SendMessage` (non-blocking) write to the DB after shutdown

**File**: `internal/bridge/bridge.go:117-129`, `cmd/.../main.go:168-176`

The non-blocking send path spawns a goroutine that calls `b.store.UpdateTaskState` after the HTTP server has shut down. `httpServer.Shutdown` doesn't track these. After `defer store.Close()` runs in `main`, the goroutine will get `sql: database is closed` errors that aren't checked. Track outstanding sends with a `WaitGroup` and drain them in shutdown.

### O5. `pluginServer.Close()` only closes the listener, not the RPC server

**File**: `internal/bridge/broker.go:225-251`

`goplugin.RPCServer.Serve(listener)` runs in a goroutine; closing the listener stops accepting new connections but doesn't drain in-flight RPCs or close `doneCh`. On a graceful shutdown, in-flight broker-published messages can race against `store.Close()` (see O4).

### O6. Broker subscription is `scion.grove.<groveId>.user.>` — too broad

**File**: `internal/bridge/bridge.go:108-114, 178-184`

This subscribes the bridge to **every** user message in the grove, not only those bound for the bridge's own user. Combined with C1, messages from other users could be reflected as agent replies. Tighten to `scion.grove.<groveId>.user.<a2a-bridge-user>.>` or move to per-task topics.

### O7. `Sender` field is parsed unauthenticated

**File**: `internal/bridge/bridge.go:588-595`

`extractAgentIDFromSender` trusts the `msg.Sender` string verbatim. If the broker does not authenticate `Sender` server-side, a spoofed message can target any agent's waiters/streams.

### O8. `subtle.ConstantTimeCompare` after the empty-key short-circuit

**File**: `internal/bridge/server.go:551`

`ConstantTimeCompare` returns 0 when lengths differ — leaking key length. Compare hashes of the keys instead, or pad/truncate before comparison.

### O9. `RateLimiter` overflow bucket is shared and never refills

**File**: `internal/bridge/ratelimit.go:75-92`

When `maxBuckets` is exceeded, all new keys collide on a single zero-rate bucket — denying every new IP indefinitely. Fine as a backstop, but document it; consider an LRU eviction instead.

### O10. `Bridge.SendMessage` blocking-mode DB updates lose errors

**File**: `internal/bridge/bridge.go:144-179`

`b.store.UpdateTaskState(taskID, …)` errors are discarded throughout `bridge.go`. State persistence is best-effort but the API contract behaves as if it's authoritative. At minimum log the failures.

### O11. Auto-provision request passes `groveSlug` as `GroveID`

**File**: `internal/bridge/bridge.go:493-507`

```go
created, err := b.hubClient.Agents().Create(ctx, &hubclient.CreateAgentRequest{
    Name:    agentSlug,
    GroveID: groveSlug,            // <-- slug, not ID
    Template: groveCfg.DefaultTemplate,
    …
})
```

Per `pkg/hubclient/agents.go:153-156`, `CreateAgentRequest.GroveID` is the canonical grove ID. Passing the user-facing slug here will work only if the Hub accepts both — verify and either normalize the slug → ID upfront, or rename the config field to make the assumption explicit.

### O12. Long-lived admin JWT for every request

**File**: `internal/identity/identity.go:83-123`, `cmd/.../main.go:97`

Every A2A client effectively acts as a single Hub admin user (`role: "admin"`, 15-minute JWT, cached). There is no per-client identity propagation to the Hub — the bridge's audit trail conflates all callers. This is acknowledged by the design, but worth flagging as a security boundary that operators must understand.

### O13. SSE `data:` framing does not handle multi-line payloads correctly

**File**: `internal/bridge/server.go:419`

```go
fmt.Fprintf(w, "data: %s\n\n", data)
```

`data` is JSON, so it is single-line in practice. But if any future event payload contains an embedded `\n`, the SSE parser on the client will split it into multiple events. Defensive: replace `\n` in `data` with `\ndata: ` per the SSE spec, or assert no newlines.

### O14. `loadSigningKey` returns whitespace-trimmed file contents

**File**: `cmd/scion-a2a-bridge/main.go:204-219`

`strings.TrimSpace` only handles leading/trailing whitespace. Embedded newlines (e.g., a wrapped base64 string) will fail decoding with a confusing "illegal base64 data" error. Strip all whitespace before decoding.

### O15. Tests use a `nil` Hub client and minter

Many handlers (auto-provision, `resolveContext`) panic when called with the test fixture. The test suite asserts the error paths but never exercises the happy path against a fake Hub. Consider adding a small `hubclient.Client` mock so the message-send path is covered end-to-end — this would have surfaced C1 and C8.

---

## Positive Feedback

- The module is **cleanly isolated** under `extras/scion-a2a-bridge` with its own `go.mod` and `replace` directive — no risk of polluting the main build, easy to revert.
- Dockerfile pins **digests for both base images**, runs as **non-root UID 1000**, and produces a small final image — exemplary container hygiene.
- JSON-RPC dispatcher uses **Go 1.22+ method-prefix routes** (`POST /…/jsonrpc`), method handlers are small and test-friendly, and the A2A error code mapping is faithful to the spec.
- Constant-time API-key compare, request-body cap (`http.MaxBytesReader(w, r.Body, 1<<20)`), and the readiness probe distinguishing DB / broker health are all good baseline practices.
- The push-notification redirect blocker (`CheckRedirect: errRedirectBlocked`) is correct and prevents the most common SSRF redirect chain — even though the DNS-rebinding hole remains (C2).
- Test coverage for the SSRF allow-list, rate-limit fallthrough, push retry/delete behavior, JSON-RPC dispatch errors, and stream-manager concurrency is appropriately broad for an MVP.
- `Bridge.activeTasks` and `streams` cleanup paths correctly delete empty map entries to avoid memory growth.

---

## Final Verdict

**Request changes — needs rework before merge.**

C1 (response cross-talk), C2/C3 (SSRF rebinding + IPv6 gaps), C4 (auth fails open), C5 (SSE killed by `WriteTimeout`), C6 (silent unsubscribe), C7 (env-expansion footgun), and C8 (cancel doesn't cancel) are blocking for any deployment that exposes the bridge to untrusted clients or runs more than one A2A caller per agent. The architecture is sound and the surface area is contained, so I expect remediation to be local and additive rather than structural — but the bridge in its current form will silently misroute responses and is exploitable for SSRF and unauthenticated access.

I'd recommend addressing C1, C2, C4, C5, C8 as a hard pre-merge bar; C3, C6, C7 as same-PR fixes; and the Observations as follow-up issues filed against the new module.

PR #219, opened 2026-05-07
fix(web): enforce asset compilation at build-time and fix runtime 404s

Jberlinsky:
This PR ensures that Scion binaries built from source have correctly compiled web assets embedded, or fail at compile-time with a helpful error message. It also fixes a runtime bug where the server would return a 404 error if the web directory was present but empty.

## Key Changes
- Compile-Time Enforcement: Added a sentinel `//go:embed` directive in `web/embed.go` for `assets/main.js`. Building the project now fails immediately if the web assets haven't been built (`make web`), preventing the creation of "broken" binaries.
- Improved Runtime Asset Detection: Modified `hasWebAssets()` in `pkg/hub/web.go` to verify the presence of `assets/main.js` before attempting to serve the SPA shell. This allows the server to gracefully fall back to a "Web UI Not Available" page even if an empty assets directory is provided via `--web-assets-dir`.
- Streamlined Onboarding: Updated the installation documentation (`install.md`) to recommend `make all`, which handles both the web build and installation to `~/.local/bin` in a single command.
- Test Setup: Updated `pkg/hub/web_test.go` to automatically provision dummy assets for web tests, ensuring CI remains green even when running without a full frontend build (e.g., using `-tags no_embed_web`).

## Why this is needed
Previously, because `web/dist/client` contained a `.gitkeep` file, the Go compiler would successfully embed an effectively empty directory. At runtime, the server would find this directory, serve the main HTML shell, and then fail to load the actual JavaScript, leading to a confusing 404 error (visible only in the browser console) and a blank page for users building from source.

## Testing Performed
- [x] Verified make build fails when assets are missing.
- [x] Verified make all successfully builds and installs a functional binary.
- [x] Verified go build -tags no_embed_web still works for CLI-only builds.
- [x] Verified all pkg/hub tests pass.

### DIFF
--- a/docs-site/src/content/docs/getting-started/install.md
+++ b/docs-site/src/content/docs/getting-started/install.md
@@ -70,7 +70,7 @@ make all
 # You can add ~/.local/bin/ to your PATH, or `mv` the binary to another directory in your PATH
 ```
 
-Alternatively, if you don't need the web UI, you can use `make build` to only compile the CLI into `./build/scion` (the web UI will show a placeholder page).
+Alternatively, if you don't need the web UI, you can build the CLI-only version using the no_embed_web tag: go build -tags no_embed_web -o ./build/scion ./cmd/scion (the web UI will show a placeholder page).
 
 To verify your installation, run:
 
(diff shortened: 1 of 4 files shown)


ptone:
rebuilding this from current main

PR #220, opened 2026-05-08
fix(hub): inject SCION_DEV_TOKEN in DispatchAgentStart

dat91:
## Summary

In dev-auth mode, agents started through the start-existing path (e.g.
`scion start <name>` on an agent previously provisioned with
`scion create`, or any broker re-dispatch routed through
`DispatchAgentStart`) reach the Hub with no usable credential. Inside
the container every hub-touching `scion` command fails with:

```
Error: authentication failed, login to hub with 'scion hub auth login'
Exit code: 1
```

`buildCreateRequest` (used by `DispatchAgentCreate` /
`DispatchAgentProvision`) injects `SCION_DEV_TOKEN = d.devAuthToken`
into `ResolvedEnv`. `DispatchAgentStart` did not. This PR adds the
matching injection so the start-existing path produces the same
in-container auth state as the create+start path.

## Reproduction (before this PR)

Same hub, same broker, same template, same operator dev token. Only
the provisioning path differs.

```sh
# Works — in-container `scion list` exits 0
scion start authtest -t pika

# Also works — restart preserves the scion-token written at first launch
scion stop authtest && scion resume authtest

# Fails — in-container `scion list` exits 1 with the auth error above
scion create devauth6 -t pika -y
scion start devauth6
```

After the fix, the third case behaves like the first.

## Code path

- `pkg/hub/httpdispatcher.go` — `buildCreateRequest` already injects
  `SCION_DEV_TOKEN` when `d.devAuthToken != ""` (around the existing
  block near line 469). `DispatchAgentStart` builds its own
  `resolvedEnv` separately and lacked the equivalent block.
- The change adds the same conditional injection at the end of
  `DispatchAgentStart`'s env-resolution section, before the debug
  summary log.

## Tests

- `TestHTTPAgentDispatcher_DispatchAgentStart_InjectsDevToken` —
  verifies `SCION_DEV_TOKEN` is set when `SetDevAuthToken` was called,
  and that existing env vars are preserved.
- `TestHTTPAgentDispatcher_DispatchAgentStart_NoDevToken` — verifies
  no `SCION_DEV_TOKEN` is set when `devAuthToken` is empty.

These mirror the existing
`TestHTTPAgentDispatcher_DispatchAgentCreate_InjectsDevToken` /
`_NoDevToken` tests.

Full `pkg/hub` test suite (1716 tests) passes locally.

## Test plan

- [x] `go test ./pkg/hub/...` passes
- [ ] Manual repro (above) flips from exit 1 to exit 0 against a
      hub running on the patched server build

### DIFF
--- a/pkg/hub/httpdispatcher.go
+++ b/pkg/hub/httpdispatcher.go
@@ -1011,6 +1011,15 @@ func (d *HTTPAgentDispatcher) DispatchAgentStart(ctx context.Context, agent *sto
 		}
 	}
 
+	// In dev-auth mode, inject the dev token so the in-container scion CLI can
+	// authenticate to the Hub. buildCreateRequest does the same for the
+	// create+start path; without it here, agents started via the start-existing
+	// path (e.g. `scion start <created-agent>`, broker re-dispatch on resume)
+	// reach the Hub with no usable credential and every hub call returns 401.
+	if d.devAuthToken != "" {
+		resolvedEnv["SCION_DEV_TOKEN"] = d.devAuthToken
+	}
+
 	if d.debug {
 		configEnvCount := 0
 		if agent.AppliedConfig != nil {
--- a/pkg/hub/httpdispatcher_test.go
+++ b/pkg/hub/httpdispatcher_test.go
@@ -1659,6 +1659,101 @@ func TestHTTPAgentDispatcher_DispatchAgentCreate_DevTokenMergesWithExistingEnv(t
 	}
 }
 
+// TestHTTPAgentDispatcher_DispatchAgentStart_InjectsDevToken verifies that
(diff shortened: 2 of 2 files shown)


dat91:
Closing this — runtime diagnostic on a real failing agent showed the in-container `~/.scion/scion-token` file contains a dev-format token (`scion_dev_<hex>`), and the CLI sends it via `X-Scion-Agent-Token` header (because `getHubClient` only does the dev-token-for-non-dev-token swap, never the inverse). The hub rejects that combination with 401. The CLI never reaches the SCION_DEV_TOKEN env fallback this PR adds, so the change is in the wrong layer.

Will open a replacement that fixes the CLI's auth-method selection — when `readAgentTokenFile()` (or `SCION_AUTH_TOKEN` env) returns a dev-format token, use `WithBearerToken` instead of `WithAgentToken`.

PR #260, opened 2026-05-29
fix(hub): support GITHUB_TOKEN secret fallback for remote template imports

### LINKED ISSUE #258: Bug: Hub unable to import templates from private repositories
### Context
When attempting to import agent templates from a private GitHub repository via either the Hub Web UI or the  POST API, the operation fails.

### Error
The Hub logs or returns the following error:
```
failed to fetch remote templates: failed to fetch GitHub folder: git fetch failed: fatal: could not read Username for 'https://github.com': No such device or address
```

### Root Cause
1. `FetchRemoteTemplate` first tries to download the repository tarball without authentication. Since the repository is private, this fails with a 404.
2. It then falls back to a sparse Git clone over HTTPS (`https://github.com/...`).
3. Since no authentication is provided and the Git process runs in a non-interactive shell, Git fails with a credentials/username prompt error (`No such device or address`).

### Proposed Solution
Support reading a project-scoped `GITHUB_TOKEN` secret from the database during remote template imports (inside `pkg/hub/template_bootstrap.go`) and pass it as the auth token parameter to `FetchRemoteTemplate` to authenticate the clone/fetch request.

chiefkarlin:
This PR resolves an authentication failure when importing remote templates from private GitHub repositories.

### Changes
* Reads project-scoped `GITHUB_TOKEN` from the secret backend (or database fallback) in `pkg/hub/template_bootstrap.go`.
* Passes the retrieved token to `FetchRemoteTemplate` to authenticate download and git operations.
* Adds an offline unit test `TestImportTemplatesFromRemote_WithProjectGithubToken` to verify the secret resolution and fallback path using a mock HTTP client.

Closes #258

### DIFF
--- a/pkg/hub/template_bootstrap.go
+++ b/pkg/hub/template_bootstrap.go
@@ -16,6 +16,7 @@ package hub
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -258,6 +259,25 @@ func (s *Server) importTemplatesFromRemote(ctx context.Context, projectID, sourc
 	if err == nil && project != nil && project.GitHubInstallationID != nil {
 		if token, _, mintErr := s.MintGitHubAppTokenForProject(ctx, project); mintErr == nil && token != "" {
 			authToken = token
+		}
+	}
+
+	// Fall back to project GITHUB_TOKEN secret if no App token minted
+	if authToken == "" {
+		if sb := s.GetSecretBackend(); sb != nil {
+			sec, secErr := sb.Get(ctx, "GITHUB_TOKEN", store.ScopeProject, projectID)
+			if secErr == nil && sec != nil && sec.Value != "" {
+				authToken = sec.Value
+			} else if secErr != nil && !errors.Is(secErr, store.ErrNotFound) {
+				s.templateLog.Warn("Failed to retrieve GITHUB_TOKEN from secret backend", "projectID", projectID, "error", secErr)
+			}
+		} else {
+			secretVal, secretErr := s.store.GetSecretValue(ctx, "GITHUB_TOKEN", store.ScopeProject, projectID)
+			if secretErr == nil && secretVal != "" {
+				authToken = secretVal
+			} else if secretErr != nil && !errors.Is(secretErr, store.ErrNotFound) {
+				s.templateLog.Warn("Failed to retrieve GITHUB_TOKEN from database", "projectID", projectID, "error", secretErr)
+			}
 		}
 	}
 
--- a/pkg/hub/template_bootstrap.go
+++ b/pkg/hub/template_bootstrap.go
@@ -25,6 +25,8 @@ import (
 	"github.com/GoogleCloudPlatform/scion/pkg/api"
 	"github.com/GoogleCloudPlatform/scion/pkg/config"
 	"github.com/GoogleCloudPlatform/scion/pkg/config/templateimport"
+	"github.com/GoogleCloudPlatform/scion/pkg/secret"
+	"github.com/GoogleCloudPlatform/scion/pkg/storage"
 	"github.com/GoogleCloudPlatform/scion/pkg/store"
 )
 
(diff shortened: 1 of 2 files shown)


ptone:
# PR #260 Review — fix(hub): support GITHUB_TOKEN secret fallback for remote template imports

- **Author:** chiefkarlin
- **State:** open
- **Base:** `main` (2c03b71f)
- **Head:** 19b0d19b
- **Files:** `pkg/hub/template_bootstrap.go` (+20), `pkg/hub/template_bootstrap_test.go` (+143)
- **Closes:** #258

## Overview

Adds a fallback in `Server.importTemplatesFromRemote` (template_bootstrap.go:454-471): if no GitHub App installation token can be minted for the project, look up a project-scoped `GITHUB_TOKEN` secret — preferring the `SecretBackend` and falling back to a direct `store.GetSecretValue` read. The resulting token is passed to `config.FetchRemoteTemplate`, which already plumbs it as a `Bearer` header on the GitHub tarball request.

## What works well

- **Tight, targeted change.** Inserts the fallback at the right place (between mint and fetch) without touching unrelated code paths. Defaulting to `""` keeps existing unauthenticated behavior intact for public repos.
- **Doesn't leak errors to callers.** `ErrNotFound` is silently swallowed (correct — "no token configured" is normal); other errors logged at WARN, not returned. Consistent with how the GitHub App mint error is silently ignored on line 449.
- **Token leakage is bounded.** `fetchGitHubTarball` only attaches the `Authorization` header on a URL it constructs itself (`https://github.com/<owner>/<repo>/archive/...`), so the project's `GITHUB_TOKEN` cannot accidentally be sent to a non-GitHub host via a crafted `sourceURL`.
- **Test exercises the meaningful path.** It seeds a real `LocalBackend`, calls the actual `importTemplatesFromRemote`, intercepts the HTTP request, and asserts the `Bearer` header is set — strong end-to-end coverage of the happy path.

## Issues / suggestions

### 1. The `sb == nil` branch is semantically suspect (template_bootstrap.go:463-469)

`store.GetSecretValue` is declared to return `encryptedValue string` (pkg/store/store.go:548-551) and the SQLite implementation reads the `encrypted_value` column (pkg/store/sqlite/sqlite.go:4946-4961). The PR uses that value verbatim as the auth token. Works today only because the "encrypted" column actually holds plaintext under the local backend (pre-existing convention also relied on by `handlers_github_app.go:245` and `server.go:870`), but it's a latent footgun: the day someone wires real at-rest encryption into the local store, this fallback silently produces garbage tokens with no error.

Consider either:
- dropping the `else` branch entirely (if `GetSecretBackend()` is nil in production the system has bigger problems), or
- adding a comment acknowledging this is plaintext-by-convention and pointing at the other call sites that share the assumption.

### 2. Test doesn't cover the fallback branch it adds

The test seeds a `LocalBackend` via `srv.SetSecretBackend(...)`, so only the `sb != nil` path is exercised. The `else` branch (lines 463-469) has zero coverage. Either delete it (see #1) or add a second test case that nils the backend.

### 3. Global mutation of `http.DefaultClient.Transport` is race-prone

```go
oldTransport := http.DefaultClient.Transport
defer func() { http.DefaultClient.Transport = oldTransport }()
http.DefaultClient.Transport = &mockRoundTripper{...}
```

If any test in this package ever calls `t.Parallel()`, this leaks across goroutines. Safer options: (a) refactor `fetchGitHubTarball` to accept an injectable `*http.Client`, or (b) use `httptest.NewServer` and rewrite the URL. If keeping the current shape, at minimum add a comment that this test must not be parallelized.

### 4. The `sync.Mutex` around `capturedAuthHeader` is dead weight

The mock is invoked from the same goroutine that calls `importTemplatesFromRemote`, so no concurrent access. Remove the mutex or document why it's there.

### 5. Style nit: scope constant

`sb.Get(ctx, "GITHUB_TOKEN", store.ScopeProject, projectID)` passes a `store.*` constant into a `secret.SecretBackend` call. They happen to share the value `"project"`, but the corresponding `secret.ScopeProject` exists (pkg/secret/secret.go:42) and would be the more honest argument. Same for the test, which already uses `secret.ScopeProject` — keep the source aligned.

### 6. Minor: no Info log on success

Both the GitHub App mint path (line 449) and this new fallback succeed silently. For first-time debugging of "why did template import auth work / not work", one `templateLog.Info("using project GITHUB_TOKEN for template import", "projectID", projectID)` on success would save someone an hour. Optional.

### 7. User-scope fallback not considered

If a user has a personal `GITHUB_TOKEN` at `ScopeUser` but the project doesn't have one, import still fails. May be intentional (project resources should resolve via project-scoped secrets) — worth a one-line code comment to make that policy explicit.

## Risk assessment

Low. The change is additive, fails closed (no token → unauthenticated request → 404 for private repos, same as today), and doesn't widen the blast radius of the token (it only reaches the GitHub tarball/git endpoints that already received App tokens). Main concerns are the dead `else` branch and its absent test coverage.

## Recommendation

Approve once #1/#2 are addressed (either delete the redundant fallback or add coverage). #3 is worth fixing if any nearby test is likely to become parallel; the rest are polish.

PR #300, opened 2026-06-03
feat(web): render agent message bodies as markdown in the Messages tab

theref:
The per-agent Messages tab rendered message bodies as plain text with a 300-char truncation and a click-to-JSON log-viewer interaction. Agent harnesses emit markdown-formatted prose (headings, tables, lists, code blocks, blockquotes), so replies were effectively unreadable.

Rework the viewer into a conversation view:

- Message bodies render as sanitized markdown (marked + DOMPurify, both already dependencies), lazy-loaded via a new shared shared/markdown.ts renderer that scion-markdown-preview now shares too.
- Oldest-first ordering with the latest at the bottom, a sticky composer, and a jump-to-latest control when scrolled up.
- Operator instructions and agent replies are visually distinguished; the raw JSON envelope is demoted to a per-message raw (</>) toggle.
- Conversational message types (assistant-reply, instruction, …) render without noisy type badges.

Rendering is harness-agnostic: it formats whatever text the message store holds, independent of which harness produced it. Frontend-only — no API, schema, or backend changes.

### DIFF
--- a/web/src/components/shared/agent-message-viewer.ts
+++ b/web/src/components/shared/agent-message-viewer.ts
@@ -17,14 +17,23 @@
 /**
  * Agent message viewer component.
  *
- * Displays structured messages from the dedicated "scion-messages" Cloud
- * Logging log. Shows message direction (sent/received), sender/recipient,
- * and provides a compose box for sending new messages with optional interrupt.
+ * Renders the conversation between an operator and an agent as a chat: the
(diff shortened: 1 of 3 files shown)


ptone:
@theref thanks for the contribution - we do require the CLA - is that something you'd be willing to sign?

ptone:
markdown rendering now directly supported in native chat

PR #566, opened 2026-07-02
fix: skip image registry rewrite for fully qualified references

ptone:
## Summary

- **RewriteImageRegistry()** now detects when an image reference already has an explicit registry hostname (first path component contains "." or ":") and skips rewriting it.
- This aligns with OCI/Docker reference semantics: a fully qualified reference like `ghcr.io/org/scion-foo:v1` means the author chose that registry deliberately.
- Makes the `image_pinned` approach from PR #425 unnecessary — the rewrite function itself now correctly detects when to skip.

## Changes

- `pkg/config/settings_v1.go`: Added explicit-registry guard in `RewriteImageRegistry()`
- `pkg/config/settings_v1_test.go`: Updated existing tests + added 3 new edge-case tests (13 total, all passing)

## Test plan

- [x] All 13 `TestRewriteImageRegistry` subtests pass
- [x] Full `pkg/config` test suite passes
- [x] Code reviewed and approved by independent reviewer

### DIFF
--- a/pkg/config/settings_v1.go
+++ b/pkg/config/settings_v1.go
@@ -210,6 +210,17 @@ func RewriteImageRegistry(fullImage, newRegistry string) string {
 		return fullImage
 	}
 
+	// If the image already has an explicit registry hostname
+	// (first path component contains "." or ":"), don't rewrite it.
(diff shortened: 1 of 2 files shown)


ptone:
replaces #425

PR #652, opened 2026-07-09
fix(server): honor explicit --dev-auth=false in workstation daemon mode

stevegeek:
## Problem

In workstation mode, `scion server start --dev-auth=false` starts the hub with **dev-auth still enabled** — it auto-generates a developer token and accepts unauthenticated admin calls. There is no way to disable dev-auth in workstation daemon mode.

## Root cause

`scion server start` (without `--foreground`) re-exec's itself as a `--foreground` child, building the child's argv in `runServerStart` (`cmd/server_daemon.go`). Workstation-defaulted bool flags were forwarded **only when true**:

```go
if enableDevAuth {
    daemonArgs = append(daemonArgs, "--dev-auth")
}
```

So an explicit `--dev-auth=false` (parsed correctly in the parent, where `Changed("dev-auth")` is true and the workstation default does not override it) is simply **omitted** from the child's argv. The child then re-runs `applyWorkstationDefaults`, which does `if !cmd.Flags().Changed("dev-auth") { enableDevAuth = true }` — and since the child never saw `--dev-auth`, it re-enables it. `loadAndReconcileConfig` writes that back as `cfg.Auth.Enabled = true`.

The same drop-on-false pattern affects every workstation-defaulted bool — `--enable-hub`, `--enable-runtime-broker`, `--enable-web`, `--auto-provide` — so an explicit subset-disable (e.g. `--enable-web=false`) is likewise silently re-enabled in the child. It is present in both the initial daemon builder and the restart reconstruct-from-flags fallback.

## Fix

Forward explicitly-set workstation-defaulted bools as `--flag=<value>` (via a new `appendDaemonBoolFlag` helper) so the disable survives into the child — the child then sees the flag as `Changed` and does not re-apply the workstation default. Flags that were not set explicitly keep the historical bare form (present only when true), so nothing changes for the common case. Applied in both arg builders.

## Testing

- New unit test (`TestAppendDaemonBoolFlagForwardsExplicitDisable`) covering: explicit false → `--flag=false`; explicit true → `--flag=true`; unset+true → bare `--flag`; unset+false → omitted.
- Verified empirically in an isolated environment (fresh HOME, spare ports): with the fix, `server start --dev-auth=false` (daemon) yields dev-auth **off** — no developer token created, unauthenticated admin rejected — matching `server start --foreground --dev-auth=false`, which was already correct (proving the daemon arg rebuild was the sole culprit).
- `go build ./...`, `gofmt`, `go vet`, and the server/workstation/daemon tests pass.

### DIFF
--- a/cmd/server_daemon.go
+++ b/cmd/server_daemon.go
@@ -52,6 +52,86 @@ func appendDaemonBoolFlag(cmd *cobra.Command, args []string, name string, val bo
 	return args
 }
 
+// buildDaemonStartArgs constructs the argv for the `server start --foreground`
+// child from the parsed server-start flags/globals. Every flag the user set
+// explicitly is forwarded (bools as --flag=<value> via appendDaemonBoolFlag,
+// string/int flags as --flag=<value> when Changed) so it survives the re-exec;
+// flags left at their defaults are omitted (workstation-defaulted bools keep
+// their historical bare-when-true form).
+func buildDaemonStartArgs(cmd *cobra.Command) []string {
+	daemonArgs := []string{"server", "start", "--foreground"}
+	if hostedMode {
+		daemonArgs = append(daemonArgs, "--hosted")
+	}
+	// Workstation-defaulted bools must be forwarded as --flag=<value> when set
+	// explicitly, so an explicit disable survives into the --foreground child.
+	// Forwarding "bare when true" alone loses --enable-web=false / --dev-auth=false
+	// etc.: the child sees the flag as unset and applyWorkstationDefaults re-enables
+	// it. See appendDaemonBoolFlag.
+	daemonArgs = appendDaemonBoolFlag(cmd, daemonArgs, "enable-hub", enableHub)
+	daemonArgs = appendDaemonBoolFlag(cmd, daemonArgs, "enable-runtime-broker", enableRuntimeBroker)
+	daemonArgs = appendDaemonBoolFlag(cmd, daemonArgs, "enable-web", enableWeb)
+	daemonArgs = appendDaemonBoolFlag(cmd, daemonArgs, "dev-auth", enableDevAuth)
+	if enableDebug {
+		daemonArgs = append(daemonArgs, "--debug")
+	}
+	daemonArgs = appendDaemonBoolFlag(cmd, daemonArgs, "auto-provide", serverAutoProvide)
+	// Remaining bool flags are not workstation-defaulted, but an explicit value
+	// set at `server start` (daemon mode) is still lost to the child's defaults
+	// unless forwarded. appendDaemonBoolFlag omits them when unset.
+	daemonArgs = appendDaemonBoolFlag(cmd, daemonArgs, "no-auto-migrate", noAutoMigrate)
+	daemonArgs = appendDaemonBoolFlag(cmd, daemonArgs, "enable-test-login", enableTestLogin)
+	daemonArgs = appendDaemonBoolFlag(cmd, daemonArgs, "simulate-remote-broker", simulateRemoteBroker)
+	daemonArgs = append(daemonArgs, fmt.Sprintf("--host=%s", hubHost))
+	if cmd.Flags().Changed("port") {
+		daemonArgs = append(daemonArgs, fmt.Sprintf("--port=%d", hubPort))
+	}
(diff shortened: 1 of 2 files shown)


stevegeek:
Yes, runServerRestart uses serverRestartCmd, which doesn’t register --enable-hub/-runtime-broker/-web/--dev-auth (those are on serverStartCmd) and never populates the globals, so Changed() is always false there and appendDaemonBoolFlag is inert.
That’s a pre-existing limitation rather than a regression; the original fallback already reduced to ["server", "start", "--foreground"] for the same reason. So we reverted the restart hunk to its original form and documented the limitation, rather than pretending to fix it here. The actual fix lives in runServerStart, whose corrected args are persisted via SaveArgs and reloaded on a normal restart. Properly reconstructing flags in the restart fallback (loading config/env) is out of scope for this change. I’ll leave it for a separate PR.

stevegeek:
/gemini review

Ok, updated as per review

ptone:
Thanks for the PR - I'm curious what auth you were wanting to use in workstation mode - were you looking to use oauth testing locally- this is for sure a bug, note that looking into this has resulted in this future clean up proposal:

https://github.com/ptone/scion/issues/396

stevegeek:
Thanks for merging! On the auth question, its really down to what Im building (which I called "Lever") to allow me to create multiagent fleets with strong containment (where host is either my local machine or a cloud instance). Heres a long winded answer:

The lever setup is to run all of scion and its agents in a VM / isolated machine so agents have no ambient authority, no host filesystem access, a locked-down egress allowlist etc. And a capability broker running on the host mediates agent credentials, tool calls and scion hub calls. So scion is the orchestration engine in the VM (the agents it runs via rootless podman) and "lever" puts around it the VM boundary plus the capabilities/permissions grants to host side resources. Also in a lever fleet there is a manager agent with whole project tree visibility and orchestration control over the fleet (the right to do so granted to it by the capability broker), and the worker agents are confined to subtrees and a narrower set of tool grants.

Here is (Claude's) description of the Lever implementation detail in relation to `dev-auth` and why we do what we do:

> We run `scion server start` in workstation/combined mode (to get the Hub API on the web port) but with `--dev-auth=false`. With dev-auth on, the hub auto-generates an admin dev token (`~/.scion/dev-token`, also exported as `SCION_DEV_TOKEN`) — ambient hub-wide admin that any co-located process able to read it inherits, which is more than we want for our agents. With it off, no agent (manager or worker) holds a hub credential at all: the controller PAT (`agent:manage,agent:attach,project:read`) lives host-side only, injected as `SCION_HUB_TOKEN` into lever's own scion client calls (the capability broker drives the hub on the manager's behalf over mTLS, and lever uses it for attach/msg/stop). The manager and workers are just scion agents, using the per-agent tokens scion mints when it dispatches them.
> 
> Bootstrapping that first PAT is the awkward part — minting one is an authenticated operation, and `dev-auth` is the only headless way to get an initial identity (OAuth + `--admin-emails` works but needs an IdP and an interactive flow). So we bring up a throwaway dev-auth hub on loopback, mint the PAT, shred the dev token, then start the real hub dev-auth-off. `hack/minttoken` already signs a user JWT offline with the hub's signing key — promoting something like it to a supported `scion hub token bootstrap` would drop the throwaway-hub dance entirely.
> 
> That's also why `--dev-auth=false` in combined mode specifically bit us: we wanted the single-web-port setup without the dev-auth default.

On #396 yeah agree. Option 3 sounds good. I think I have a branch for the Option 1 you mention if you would like a PR for that approach (https://github.com/GoogleCloudPlatform/scion/compare/main...stevegeek:scion:fix/server-restart-reconstruct-daemon-args?expand=1)

ptone:
Thanks for the reply - I understand a bit more about now how you are using the accesstoken (PAT) from Scion.

It might be interesting to discuss what it would take to the features of agent-auth and permissions already in scion and enable enough control for you to just run scion in "lever mode" to be as locked down as you are seeking.  I think it could be a matter of mostly making sure agents started have minimal/no permissions from their minted tokens.

If this sounds interesting hit me up in email (not always great at seeing, be sure to put scion in subject line - : ptone at google.com - or DM on X)

stevegeek:
Thank you, will have a think and ping you an email

PR #793, opened 2026-07-18
harness/antigravity: read SCION_THINKING_LEVEL instead of AGY_THINKING_LEVEL

ptone:
Updates antigravity to read the standard auto-injected SCION_THINKING_LEVEL env var instead of the legacy AGY_THINKING_LEVEL. SCION_THINKING_LEVEL is injected from AgentAppliedConfig.ThinkingLevel (see companion PR adding --thinking-level flag)


ptone:
We may need to ensure that the original AGY env car is still a har ess native var.  And not mess with it. Or defer to it.

PR #881, opened 2026-07-28
scion-messaging: code-traced message length limit

ptone:
Replaces the softer wording from #875 with exact code-path data (handler, rune count, dead-code note)


ptone:
related to the section added on ## Inbound Message Types

Waiting for input may absolutely be directed to a parent agent who is responsible for answering the agent's question

PR #953, opened 2026-08-02
feat(harness): auto-detect and expose listening ports in agent containers

ptone:
Adds automatic port detection and exposure for agent containers. When enabled via SCION_AUTO_EXPOSE_PORTS=true, a reconciliation loop in sciontool init scans /proc/net/tcp and /proc/net/tcp6 for LISTEN sockets and auto-registers/deregisters them with the Hub.

Key design:
- New pkg/sciontool/autoexpose/ package (scanner, reconciler, config)
- Diff-based reconciliation on configurable ticker (default 3s)
- Reuses existing Hub client library (RegisterPort/DeletePort/ListPorts) -- no new API endpoints
- Integration as background goroutine in init.go alongside heartbeat and port-forward tunnel

Security controls (all required, all implemented):
- Opt-in disabled by default (SCION_AUTO_EXPOSE_PORTS env var)
- Allowlist filter mode by default (empty allowlist = nothing exposed)
- Denied infrastructure ports (8080, 9810, 18380) enforced client-side
- Minimum port threshold (default 1024)
- Auto-exposed ports labeled "auto-scan" to distinguish from manual
- Never auto-unexpose manually-exposed ports

Files: 7 (new package + init.go integration), 30 tests
Two independent fork review rounds, both approved.

Ref: ptone/scion#699

### DIFF
--- a/cmd/message_test.go
+++ b/cmd/message_test.go
@@ -618,7 +618,7 @@ func TestSendOutboundMessageViaHub(t *testing.T) {
 		case r.URL.Path == "/healthz" && r.Method == http.MethodGet:
 			_ = json.NewEncoder(w).Encode(map[string]interface{}{"status": "ok"})
 		case r.Method == http.MethodPost && (r.URL.Path == "/api/v1/projects/"+projectID+"/agents/my-agent/outbound-message" ||
-				r.URL.Path == "/api/v1/groves/"+projectID+"/agents/my-agent/outbound-message"):
+			r.URL.Path == "/api/v1/groves/"+projectID+"/agents/my-agent/outbound-message"):
 			var msg hubclient.OutboundMessageRequest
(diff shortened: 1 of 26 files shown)


ptone:
A few other tweaks to add: I want to double-check that expose port can be called repeatedly (idempotent).   the cmd to list ports in sciontool should declare whether auto-listing is enabled, and the help text for any of the commands should return this state as well.  We need to clarify / decide whether unexpose is an error, or is the client side keeping track of all ports in memory state?

For the auto-deny of infrastructure ports, there are two important caveats.  1 - 8080 may be a loopback port, but served through a load-balancer where the hub API is exposed on a different port, making the path based forwarding valid.

because we are using a reverse tunnel - I don't think those host side ports are ever invalid

PR #1006, opened 2026-08-03
fix(autoexpose): suggest sharing proxy URL with collaborating users

ptone:
## Summary
- Updates the auto-expose port notification to suggest agents share the proxy URL with collaborating users

## Changes
- pkg/sciontool/autoexpose/reconciler.go - one-line message text update

## Test plan
- [ ] go build ./pkg/sciontool/... compiles
- [ ] Auto-exposed port notification includes sharing suggestion


ptone:
Rather than "Run 'sciontool port list' for the full proxy URL."" - we should just provide the full URL in this message

PR #1227, opened 2026-08-20
fix(secret): correct inverted scope precedence in Resolve() and dedup

### LINKED ISSUE #1226: Secret scope precedence inverted in Resolve() and DeduplicateByTarget — broker/project can silently outrank user
## Summary

Three independent places in `pkg/secret` rank secret scopes in an order that contradicts the documented precedence (`runtime_broker < hub < project < user`, lowest first — see the `envScopePrecedence` comment in `pkg/hub/httpdispatcher.go`). Each was implementing the *inverse* order, so `runtime_broker` (meant to be weakest) or `project` (meant to be weaker than `user`) can silently win over a stronger scope's value for the same secret.

## Where

1. **`pkg/secret/localbackend.go`**, `LocalBackend.Resolve()` — builds its scope list as `[hub, user, project, broker]` and does `merged[s.Key] = ...` per scope in that order (last write wins), making **broker the strongest scope** instead of the weakest.
2. **`pkg/secret/gcpbackend.go`**, `GCPBackend.Resolve()` — identical bug, same scope list, same last-write-wins pattern.
3. **`pkg/secret/secret.go`**, `scopePrecedence()` (used by `DeduplicateByTarget`, which resolves collisions between *differently-named* secrets that happen to share an injection target) — ranks `hub=1, user=2, project=3, runtime_broker=4` (higher wins), again putting broker highest and user second-lowest.

## Impact

Any project with the same secret key (or same injection target) set at more than one scope can get the wrong value injected into an agent container, with no error — it just silently uses the weaker-scope value. This is easy to hit in practice: our repro was `GITHUB_TOKEN`, which needed setting at project scope for Hub-mode's git-clone-at-agent-start to work (`scion hub env set --project --always --secret GITHUB_TOKEN=...`). A leftover `runtime_broker`-scoped copy from earlier debugging silently won over the project-scoped one, and the resulting failure mode was actively misleading — the container's clone step reported `no GITHUB_TOKEN secret configured`, which reads as "the secret doesn't exist" rather than "a different scope's value was used instead."

## Confirmation

Root-caused by patching temporary logging into a local build and tracing the resolved value through the dispatch/broker pipeline hop by hop — confirmed live that `secretBackend.Resolve()` returned the `runtime_broker`-scoped, `as_needed`-mode copy instead of the `project`-scoped, `always`-mode one, with both present simultaneously.

There's also existing test coverage that encodes the bug as expected behavior rather than catching it:
- `TestLocalBackend_ResolveBrokerOverride` in `pkg/secret/localbackend_test.go` explicitly asserts a broker-scoped secret should override a user-scoped one.
- `TestLocalBackend_Resolve`, `TestLocalBackend_ResolveDuplicateTargetAcrossScopes`, `TestLocalBackend_ResolveDuplicateEnvTargetAcrossScopes` (same file) and `TestGCPBackend_Resolve` (`gcpbackend_test.go`) all assert `project` wins over `user`, which is also backwards under the documented order.

## Fix

Opening a PR that:
- Reorders the scope list in both `Resolve()` implementations to `[runtime_broker, hub, project, user]`, matching `envScopePrecedence`.
- Corrects `scopePrecedence()`'s ranks to `runtime_broker=1, hub=2, project=3, user=4`.
- Updates the five tests above to assert the correct (documented) precedence instead of the inverted one, since they were asserting the bug.

All of `pkg/secret`'s existing tests pass after the fix (with the five corrected).

iJuanPablo:
## Summary

Fixes #1226 — secret scope precedence was inverted in three independent places, all putting `runtime_broker` (meant to be the weakest scope) ahead of `project` and `user`, and `project` ahead of `user`. The documented/intended order (see `envScopePrecedence` in `pkg/hub/httpdispatcher.go`) is:

```
runtime_broker < hub < project < user   (lowest precedence first)
```

## What was wrong

- `LocalBackend.Resolve()` (`pkg/secret/localbackend.go`) built its scope list as `[hub, user, project, broker]` and merged last-write-wins, making broker strongest.
- `GCPBackend.Resolve()` (`pkg/secret/gcpbackend.go`) — identical bug.
- `scopePrecedence()` (`pkg/secret/secret.go`, used by `DeduplicateByTarget` to resolve collisions between differently-named secrets sharing an injection target) ranked `hub=1, user=2, project=3, runtime_broker=4` (higher wins) — same inversion.

In practice this means a secret set at a weaker scope (e.g. `runtime_broker`, left over from a debugging session) can silently shadow the same key at a stronger scope (e.g. `project`), with no error — whichever scope happened to be enumerated last in the old order wins, regardless of which one the caller actually intended to take effect. Full repro and impact in #1226.

## Fix

- Reordered the scope list in both `Resolve()` implementations to `[runtime_broker, hub, project, user]`, matching `envScopePrecedence`.
- Corrected `scopePrecedence()`'s numeric ranks to `runtime_broker=1, hub=2, project=3, user=4`.
- Added a comment on each cross-referencing the other two implementations, since this is the second time the same ordering has needed fixing in three places independently (see the existing precedence-history comment above `envScopePrecedence` in `httpdispatcher.go`) — the comments try to make it easy to find all three the next time this needs to change.

## Tests

Five existing tests asserted the *inverted* precedence as expected behavior — they were passing because they were testing the bug, not despite it:

- `TestLocalBackend_ResolveBrokerOverride` → renamed `TestLocalBackend_ResolveUserOverridesBroker`, assertion flipped
- `TestLocalBackend_Resolve`, `TestLocalBackend_ResolveDuplicateTargetAcrossScopes`, `TestLocalBackend_ResolveDuplicateEnvTargetAcrossScopes` (`pkg/secret/localbackend_test.go`) — assertions flipped from "project wins" to "user wins"
- `TestGCPBackend_Resolve` (`pkg/secret/gcpbackend_test.go`) — same
- `TestResolveSecrets`, `TestResolveSecrets_WithBackend` (`pkg/hub/resolve_secrets_test.go`) — same, at the `dispatcher.resolveSecrets()` level

All updated to assert the documented precedence instead. `go test ./pkg/secret/... ./pkg/hub/... ./pkg/store/...` passes (five pre-existing, unrelated failures on this machine — `TestClassifyPath_*`/`TestFS*` — fail identically on an unpatched checkout at the same commit; looks like a macOS path-assumption issue in those tests, not something this PR touches).

## How this was found

Root-caused live while debugging why a project-scoped `GITHUB_TOKEN` (set with `--always` injection mode) wasn't reaching an agent container for Hub-mode's git-clone-at-start step, despite being stored correctly — a stale `runtime_broker`-scoped copy from an earlier debugging attempt was silently winning instead. Confirmed by patching temporary logging into a local build and tracing the resolved value hop by hop through the dispatch pipeline. Full writeup in #1226.

### DIFF
--- a/pkg/hub/resolve_secrets_test.go
+++ b/pkg/hub/resolve_secrets_test.go
@@ -112,16 +112,17 @@ func TestResolveSecrets(t *testing.T) {
 		byName[rs.Name] = rs
 	}
 
-	// API_KEY should be overridden by project scope
+	// API_KEY: user is the strongest scope (runtime_broker < hub < project
+	// < user), so the user-scoped value wins over the project-scoped one.
 	apiKey, ok := byName["API_KEY"]
 	if !ok {
 		t.Fatal("expected API_KEY in resolved secrets")
 	}
-	if apiKey.Value != "project-api-key" {
-		t.Errorf("expected API_KEY value from project scope %q, got %q", "project-api-key", apiKey.Value)
+	if apiKey.Value != "user-api-key" {
+		t.Errorf("expected API_KEY value from user scope %q, got %q", "user-api-key", apiKey.Value)
 	}
-	if apiKey.Source != store.ScopeProject {
-		t.Errorf("expected API_KEY source %q, got %q", store.ScopeProject, apiKey.Source)
+	if apiKey.Source != store.ScopeUser {
+		t.Errorf("expected API_KEY source %q, got %q", store.ScopeUser, apiKey.Source)
 	}
 
 	// DB_PASS should come from project scope
@@ -228,16 +229,16 @@ func TestResolveSecrets_WithBackend(t *testing.T) {
 		byName[rs.Name] = rs
 	}
 
-	// API_KEY should be overridden by project scope
+	// API_KEY: user is the strongest scope, so it wins over project.
 	apiKey, ok := byName["API_KEY"]
 	if !ok {
 		t.Fatal("expected API_KEY in resolved secrets")
 	}
-	if apiKey.Value != "project-api-key" {
-		t.Errorf("expected API_KEY value %q, got %q", "project-api-key", apiKey.Value)
+	if apiKey.Value != "user-api-key" {
+		t.Errorf("expected API_KEY value %q, got %q", "user-api-key", apiKey.Value)
 	}
-	if apiKey.Source != store.ScopeProject {
-		t.Errorf("expected API_KEY source %q, got %q", store.ScopeProject, apiKey.Source)
+	if apiKey.Source != store.ScopeUser {
+		t.Errorf("expected API_KEY source %q, got %q", store.ScopeUser, apiKey.Source)
 	}
 
 	// DB_PASS target should be preserved
--- a/pkg/secret/gcpbackend.go
+++ b/pkg/secret/gcpbackend.go
@@ -276,17 +276,27 @@ func (b *GCPBackend) Resolve(ctx context.Context, userID, projectID, brokerID st
 		scopeID string
 	}
 
+	// Scope precedence, lowest first: runtime_broker < hub < project < user.
+	// Later entries in this slice overwrite earlier ones in the merge loop
+	// below, so this order must match envScopePrecedence in
+	// pkg/hub/httpdispatcher.go — broker is the most infrastructural and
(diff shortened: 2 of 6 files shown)


ptone:
Thanks - a few small nits, but I can address these in followup - we are starting a bunch of work to harden and tighten authz, permisisons, and policy - so you can expect more improvements in these areas

PR #1230, opened 2026-08-20
fix(hub): implement implicit recipient fallback for agent outbound messages

### LINKED ISSUE #1229: Outbound messages with no explicit recipient fail with 400, despite documented implicit-creator fallback
## Summary

`handleAgentOutboundMessage` (`pkg/hub/handlers_agent_messaging.go`) is
documented — both in its doc comment and in an inline comment above the
recipient-resolution block — as defaulting the recipient to the agent's
creator when none is explicitly specified:

> "The recipient defaults to the agent's creator when not explicitly
> specified."

> `// Resolve recipient: explicit takes precedence; implicit defaults to
> agent creator.`

No such fallback is implemented. When both `recipient` and `recipient_id`
are empty, the handler goes straight to:

```go
if recipientID == "" && recipient == "" {
    ValidationError(w, "recipient is required — specify a user with 'user:<name>' or 'user:<email>'", nil)
    return
}
```

## Impact

This breaks every caller that relies on implicit recipient resolution — most
notably the assistant-reply Stop hook
(`pkg/sciontool/hooks/handlers/hub.go`) that auto-forwards an agent's final
reply via `SendOutboundMessage()`. That call only ever sets `Msg`, `Type`,
`Visibility`, and `Metadata` — never `Recipient`/`RecipientID` — so every
automatic forward of an agent's reply is rejected with a 400. This affects
both the web dashboard's "Messages" tab and any plugin built on top of
outbound messages (e.g. the Telegram plugin): the agent's reply is captured
successfully by the hook, then silently fails to deliver.

## Repro

1. Start an agent, send it an inbound message (e.g. via Telegram plugin or
   `scion message`).
2. Let it reply. The Stop hook fires `SendOutboundMessage()` with no
   recipient.
3. Server logs a `400 validation_error: "recipient is required..."`
   immediately after the hook's message-capture step. The reply never
   reaches any consumer (dashboard Messages tab, Telegram, etc).

## Fix

PR incoming: implement the documented fallback by resolving
`agent.CreatedBy` (falling back to `agent.OwnerID`) via `store.GetUser` when
no recipient is supplied at all, before returning the validation error.

iJuanPablo:
Fixes #1229

## Summary

`handleAgentOutboundMessage`'s doc comment and an inline comment both
promise that outbound messages default their recipient to the agent's
creator when none is explicitly specified, but that fallback was never
implemented — the function returned a 400 `"recipient is required"`
immediately whenever both `recipient` and `recipient_id` were empty.

This silently broke every automatic reply-forwarding path, most notably the
assistant-reply Stop hook (`pkg/sciontool/hooks/handlers/hub.go`) used by
both the web dashboard's Messages tab and the Telegram plugin: an agent's
final reply is captured by the hook and forwarded via
`SendOutboundMessage()`, which never sets `Recipient`/`RecipientID`, so the
Hub rejects it with 400 and the reply never reaches the user.

## Fix

When no recipient is given at all, resolve `agent.CreatedBy` (falling back
to `agent.OwnerID` for agents created by automation with no creator) via
`store.GetUser`, mirroring the existing explicit-recipient resolution logic
just above it (email lookup / display-name search / `"user:<name>"`
formatting).

## Testing

- Added `TestOutboundMessage_ImplicitRecipientDefaultsToCreator` — creates
  an agent with `CreatedBy` set and no recipient in the request, asserts a
  2xx response instead of the previous 400.
- `go build ./pkg/hub/...` — clean.
- `go test ./pkg/hub/... -run TestOutboundMessage` — all pass, including the
  new test and the existing `TestOutboundMessage_UnknownRecipient` (explicit
  bad recipient still correctly rejected).
- Full `go test ./pkg/hub/...` — same 5 pre-existing failures on unpatched
  `main` (`TestClassifyPath_ManagedPath`, `TestClassifyPath_ManagedLegacyGroves`,
  `TestFSValidatePath_ManagedOverlap`, `TestFSList_HomeDir`,
  `TestFSList_DefaultsToHome` — unrelated `fs_safety`/filesystem tests,
  verified to fail identically with this change stashed out).

### DIFF
--- a/pkg/hub/handlers_agent_messaging.go
+++ b/pkg/hub/handlers_agent_messaging.go
@@ -162,6 +162,29 @@ func (s *Server) handleAgentOutboundMessage(w http.ResponseWriter, r *http.Reque
 		}
 	}
 
+	// Implicit default: no recipient specified at all — fall back to the
+	// agent's creator (falling back further to its owner if creator is
+	// unset, e.g. agents created by automation). This is what lets an agent
+	// just call "send this reply back" without threading a recipient
+	// through every message-sending path (assistant-reply hooks, Telegram
+	// relay, etc.).
+	if recipientID == "" && recipient == "" {
+		creatorID := agent.CreatedBy
+		if creatorID == "" {
+			creatorID = agent.OwnerID
+		}
+		if creatorID != "" {
+			if u, err := s.store.GetUser(ctx, creatorID); err == nil {
+				recipientID = u.ID
+				name := u.DisplayName
+				if name == "" {
+					name = u.Email
+				}
+				recipient = "user:" + name
+			}
+		}
+	}
+
 	if recipientID == "" && recipient == "" {
 		ValidationError(w, "recipient is required — specify a user with 'user:<name>' or 'user:<email>'", nil)
 		return
--- a/pkg/hub/handlers_agent_messaging.go
+++ b/pkg/hub/handlers_agent_messaging.go
@@ -174,13 +174,24 @@ func (s *Server) handleAgentOutboundMessage(w http.ResponseWriter, r *http.Reque
 			creatorID = agent.OwnerID
 		}
 		if creatorID != "" {
-			if u, err := s.store.GetUser(ctx, creatorID); err == nil {
+			u, err := s.store.GetUser(ctx, creatorID)
+			switch {
+			case err == nil:
 				recipientID = u.ID
 				name := u.DisplayName
 				if name == "" {
 					name = u.Email
 				}
 				recipient = "user:" + name
+			case errors.Is(err, store.ErrNotFound):
+				// Creator/owner record no longer exists (e.g. deleted user).
+				// Fall through to the "recipient is required" response below.
+			default:
+				// A real backend error (DB down, etc.) shouldn't be reported
+				// as a 400 validation error — that would mask a transient
+				// failure as a client mistake.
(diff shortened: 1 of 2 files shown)


ptone:
Thanks so much for your contribution - there are a couple reasons this may not be best way forward.

1) explicit better than implicit - while the docs/contract may indicate the implicit fallback as a plan - I think sending errors early on in agent sessions builds better "habits" and results in more consistent tool use --  I've seen agents develop "lazy/sloppy" habits in using messaging if they don't get errors on mis-use early on (there are places where this needs tightening.

2) the other reason - is now with chat, and the native chat becoming more prominent, I want a channel+thread to be a valid "recipient" and get across the idea that you are posting to a chat space, not to a person.  

If there is a specific creator recipient - this makes a channel target a little more ambiguous - WDYT?

which chat integration are you using the most?

stevegeek:
I've been trying to work through this issue too, and understand the desire for it not to be implicit. I worked with Claude to try to solve this differently and have a possible proposal in #1259: rather than defaulting the recipient, it records the thread alongside the channel in the existing reply-affinity context, so an agent's untagged reply lands back in the conversation it answers, ie addressing the *where* rather than the *who*.

ptone:
See the proposed plan at https://github.com/GoogleCloudPlatform/scion/discussions/1264

ptone:
supersceded by https://github.com/GoogleCloudPlatform/scion/discussions/1264

PR #1239, opened 2026-08-20
Don't override a hub-managed agent's real token with dev auth on localhost

dea-rour:
## Summary

`createHubClient()` in `pkg/hubsync/sync.go` (and a duplicated copy,
`getHubClient()`, in `cmd/hub.go`, plus the display-only `getAuthInfo()`)
prefers dev auth over a non-dev `scion-token` whenever the Hub endpoint is
localhost, specifically to avoid a stale agent token left over on a
developer's machine from a previous *remote* hub connection.

That exception fires unconditionally — including inside a container the
Runtime Broker itself started. Workstation mode (Hub + Broker + Dashboard on
one machine, per the docs) always runs the Hub on `localhost`, so *every*
hub-managed agent in workstation mode hits this branch. Inside that
container the `scion-token` is freshly minted for this exact Hub, not stale
at all — but it gets silently swapped for dev auth anyway.

Dev auth resolves to a generic/dev identity, not a specific agent. Any Hub
endpoint that requires the caller to *be* a particular agent breaks under
it — concretely, `POST /api/v1/agents/{id}/outbound-message` (used by
`scion message user:<x>`, an agent's only way to reach a human's inbox
directly) 401s with "Agents can only send outbound messages as themselves",
surfaced to the CLI user as the generic `authentication failed, login to
hub with 'scion hub auth login'`.

Net effect: no hub-managed agent running against a workstation-mode Hub
could ever message a user directly via `scion message` — a full feature
gap for any agent/harness that doesn't have its own out-of-band relay to
the message broker (discovered running an OpenCode-harnessed agent, which
has no such native relay, alongside a Claude Code agent that does).

## Fix

Gate the localhost dev-auth-preference exception behind
`!config.IsHubManagedAgent()` — an existing helper keyed off
`SCION_AGENT_ID`, which the Runtime Broker sets when it starts an agent
container — in both `createHubClient` and `getHubClient`, and apply the
same condition in `getAuthInfo` so `scion hub auth status` reports
accurately from inside an agent container too.

The original developer-workstation scenario this exception was written for
is untouched: it only ever applied when `SCION_AGENT_ID` isn't set, which
is true for any normal CLI invocation outside an agent container.

## Test plan

- [x] `go build ./...` — clean
- [x] `go vet ./cmd/... ./pkg/hubsync/... ./pkg/config/...` — clean
- [x] `go test ./cmd/... ./pkg/hubsync/...` — all passing
- [x] Added `TestGetAuthInfo_HubManagedAgentUsesRealTokenOnLocalhost` and
      `TestCreateHubClient_HubManagedAgentUsesRealTokenOnLocalhost` —
      regression coverage for the fixed case (agent token used, not dev
      auth, when `SCION_AGENT_ID` is set on a localhost endpoint)
- [x] Fixed `TestGetAuthInfo_DevAuthPreferredOverStaleAgentTokenOnLocalhost`
      to explicitly clear `SCION_AGENT_ID` — a latent test-hermeticity gap
      the fix exposed (the existing suite happened to always run outside
      any agent context, so nothing had previously needed this cleared)
- [x] Verified live against a real workstation-mode deployment: an agent
      container that previously 401'd running `scion message
      user:<name> --channel telegram "..."` from inside itself

### DIFF
--- a/cmd/hub.go
+++ b/cmd/hub.go
@@ -452,6 +452,11 @@ func getAuthInfo(settings *config.Settings, endpoint string) authInfo {
 	// When the endpoint is localhost and dev auth is available, prefer dev auth
 	// over a non-dev agent token — the scion-token may be stale from a previous
 	// remote hub connection while the dev-token was written by the running local server.
+	// Not for hub-managed agents (config.IsHubManagedAgent()): inside a
+	// broker-started container the scion-token is freshly minted for this Hub,
+	// not stale, and dev auth doesn't carry the per-agent identity some
+	// endpoints require — see the comment on createHubClient in
+	// pkg/hubsync/sync.go.
 	if token := readAgentTokenFile(); token != "" {
 		if apiclient.IsDevToken(token) {
 			info.Method = "Agent token (dev)"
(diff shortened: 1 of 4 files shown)


ptone:
Thanks for the contribution - this looks correct, can you click through and sign the CLA?

PR #1263, opened 2026-08-23
fix(hub): serialize message timestamps as UTC

stevegeek:
One conversation mixes Z-suffixed and offset-suffixed RFC3339Nano timestamps: the chat send path stamps `CreatedAt` in UTC (`handlers_chat_v2.go`), while nine other message-writing sites — agent messaging, broker delivery, the notification inbox mirror — stamp server-local time.

Mixed formats are the defect regardless of any particular parser. Observed in the field on iOS Safari: agent replies displayed skewed by exactly the server's UTC offset while the viewer's own Z-stamped messages in the same thread displayed correctly — consistent with the offset being ignored on parse. Current JavaScriptCore parses these forms correctly, so the affected engine version wasn't pinned down; historical WebKit rejected long-fraction-plus-offset timestamps outright (moment#1200, date-fns#1924). Serializing UTC everywhere removes the class rather than arguing with parsers.

## The change — three layers, so the property holds structurally

- The nine local stamps now match the chat path's `.UTC()`.
- `CreateMessage` normalizes `CreatedAt` at the single point every message passes through, covering current and future callers (including the ent schema default, which is local).
- The read side (`entMessageToStore`) normalizes too. This is what fixes **Postgres** — pgx re-localizes `timestamptz` to process-local on scan regardless of what was stamped — and makes **historical rows** serialize as UTC instead of waiting for time to heal them.

## One-time boundary worth knowing

Rows written before the fix keep local-zone storage; in SQLite's TEXT ordering they interleave with new rows inside a stratum as wide as the server's offset. The chat client re-sorts fetched pages by epoch, and the same misordering already existed continuously between chat sends and agent stamps — so this converts a permanent skew into a one-time artifact.

## Testing

- `go test ./pkg/hub/ -run TestOutboundMessageTimestampIsUTC` — posts an outbound message and asserts the stored row reads back with zero offset, through the real store.
- The test **pins a non-UTC local zone for its duration**: on a UTC host local time *is* UTC and the assertion could never fire — which is exactly what CI runners are. Verified under `TZ=UTC`: the test fails against the reverted implementation (offset 7200s detected) and passes against the fix.
- `go build ./...` clean.
- **Note:** six tests fail on `main` in `pkg/hub` (`TestBypassAgents_404Before403Ordering`, `TestClassifyPath_*`, `TestFSValidatePath_ManagedOverlap`, `TestFSList_*`) — pre-existing, environment-dependent, identical on a clean checkout.

## Verified live

Deployed against a real hub with a UTC+3 viewer talking to a UTC+2 server: new agent replies carry `Z` and display at the correct local time; the viewer's exact symptom (replies two hours late beside correctly-timed own messages) is gone.

### DIFF
--- a/pkg/hub/handlers_agent_messaging.go
+++ b/pkg/hub/handlers_agent_messaging.go
@@ -228,7 +228,7 @@ func (s *Server) handleAgentOutboundMessage(w http.ResponseWriter, r *http.Reque
 		Channel:     req.Channel,
 		ThreadID:    req.ThreadID,
 		Visibility:  req.Visibility,
-		CreatedAt:   time.Now(),
+		CreatedAt:   time.Now().UTC(),
 	}
 
 	// Build a structured message for external dispatch paths.
@@ -709,7 +709,7 @@ func (s *Server) handleAgentMessage(w http.ResponseWriter, r *http.Request, id s
 			Channel:       structuredMsg.Channel,
 			ThreadID:      structuredMsg.ThreadID,
 			DispatchState: store.MessageDispatchDispatched,
-			CreatedAt:     time.Now(),
+			CreatedAt:     time.Now().UTC(),
 		}
 		// Propagate GroupID from metadata so CLI-originated group[] messages
 		// preserve correlation in the store.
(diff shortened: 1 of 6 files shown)


stevegeek:
@ptone Was this close intentional? It came via d6fd3204 (“security: add caller verification to broker-scoped handlers”, #1296), which carries a Closes #1263 — but that commit only touches handlers_runtime_brokers.go, so I couldn’t tell whether the reference was meant for a different number, or whether this is being folded into the wider messaging work.
As far as I can see the change isn’t on main yet — handlers_agent_messaging.go still has 8 time.Now() calls with 1 .UTC()

PR #1371, opened 2026-08-29
feat: messaging authorization with per-agent message modes (D1-D10)

ptone:
## Summary

Implements messaging authorization design (D1-D10). Messaging becomes a first-class axis with per-agent message modes enforced at a single choke point across all four ingresses.

Phase 1: Schema, registry, role seeds
Phase 2: authorizeAgentMessage choke point
Phase 3: Ingress conversion (chat v2, broadcast, broker)
Phase 4: set_message_mode endpoint with cascade + audit
Phase 5: Documentation


ptone:
remove the project log files from .design need only the primary design docs for this work.

PR #1397, opened 2026-08-29
fix(harness): opencode vertex-ai provider routing fails due to GITHUB_TOKEN collision

ptone:
## Summary
Fix 5 compounding bugs that prevent VertexAI routing in the opencode harness:

1. Strip GITHUB_TOKEN from opencode child process — prevents Copilot auto-detection from hijacking provider selection
2. Add VERTEXAI_LOCATION to env overlay — opencode VertexAI client requires this exact var name
3. Fix config filename to .opencode.json — viper reads dot-prefixed name
4. Fix MCP config key to mcpServers — matches Go config struct
5. Write vertex provider config with copilot apiKey override and Gemini models

Relates to ptone/scion#1366

## Test plan
- Live-tested: opencode agent with vertex-ai auth successfully ran Gemini model and executed tools
- Python syntax validation passed
- No Go code modified


ptone:
We should not simply globally skip the GitHub token out of the open code session, as it is also used for Git operations by command line tools. Instead, we should change the way that `openno.gethowtheart` method is used and use a different variable name for that, so the auto detection

PR #1607, opened 2026-09-15
feat: add annotation prefix guard for scion.io/ to scion.dev/ migration

ptone:
Add annotation prefix guard for scion.io/ to scion.dev/ migration. CI check scans Go files, YAML manifests, and Helm templates for the legacy scion.io/ annotation prefix and flags them for migration.

Ref: ptone/scion#1207


ptone:
This needs more work - as if we are going to actually moved to a semi-managed domain, it would be scion-ai.dev

PR #1618, opened 2026-09-15
fix: upgrade gh CLI from stale apt package to pinned v2.97.0 release

ptone:
Upgrade gh CLI from stale unpinned apt package to pinned v2.97.0 release download. The apt repo installs an outdated version; this pins a specific release for reproducibility.

Ref: ptone/scion#677


ptone:
Should be now 2.100

https://github.com/cli/cli/releases

PR #1621, opened 2026-09-15
feat: add agent-facing Discord tools via BrokerQuery

ptone:
Add agent-facing Discord tools via BrokerQuery plugin interface. Enables agents to list channels/threads, set defaults, read channel history, and send DMs through the broker integration layer.

Ref: ptone/scion#675


ptone:
being supersceded by a more general conversations tool

PR #1676, opened 2026-09-16
fix(a2a-bridge): preserve artifact reply text, fix multi-turn subscriptions, raise HTTP write timeout

zahmedsafeer-hub:
## Summary
Fixes single-turn and multi-turn message delivery in `scion-a2a-bridge`:

1. **`internal/bridge/executor.go`**: `TranslateScionToA2A` emits both an `Artifact` and a `Message` event on `assistant-reply`, with the artifact appended first. `taskEventToSDKEvent` previously returned `nil` message on `case "artifact":`, resulting in empty `TASK_STATE_COMPLETED` payloads. This patch extracts text parts from `TaskArtifactUpdate` and attaches them to `TaskStatusUpdateEvent`.
2. **`internal/bridge/broker.go`**: Removes the `if already { return nil }` short-circuit in `BrokerServer.RequestSubscription` so repeat subscription requests always forward to host callbacks for Turn 2+ and recover cleanly if the initial call occurs before host callbacks are ready.
3. **`cmd/scion-a2a-bridge/main.go`**: Raises HTTP `WriteTimeout` from `30s` to `300s` so synchronous `SendMessage` calls do not close prematurely during longer LLM reasoning turns.
4. **`README.md`**: Updates Section 4 YAML snippet to use `server.plugins.broker.a2a-bridge` matching `pkg/config/settings_v1.go`.


ptone:
## Code Review: PR #1676

### Critical — Compilation Errors (must fix before merge)

**`extras/scion-a2a-bridge/internal/bridge/executor.go`**, artifact case (~lines 228-238):

Two type mismatches prevent this code from compiling:

1. **Slice type:** `var artParts []a2a.Part` should be `var artParts []*a2a.Part` — `a2a.NewTextPart()` returns `*a2a.Part`. The existing `message` case on line 200 uses `[]*a2a.Part`.

2. **Wrong argument to `NewMessageForTask`:** The call passes `execCtx.TaskID` (type `a2a.TaskID`) but `a2a.NewMessageForTask` expects `*a2asrv.ExecutorContext` as its second argument. The `message` case on line 214 shows the correct pattern: `a2a.NewMessageForTask(a2a.MessageRoleAgent, execCtx, artParts...)`.

Suggested fix:
```go
case "artifact":
    var au TaskArtifactUpdate
    if err := json.Unmarshal(ev.Payload, &au); err != nil {
        return nil, fmt.Errorf("unmarshal artifact event: %w", err)
    }
    var artParts []*a2a.Part
    for _, p := range au.Artifact.Parts {
        if p.Text != "" {
            artParts = append(artParts, a2a.NewTextPart(p.Text))
        }
    }
    if len(artParts) == 0 {
        return a2a.NewStatusUpdateEvent(execCtx, a2a.TaskStateCompleted, nil), nil
    }
    artMsg := a2a.NewMessageForTask(a2a.MessageRoleAgent, execCtx, artParts...)
    return a2a.NewStatusUpdateEvent(execCtx, a2a.TaskStateCompleted, artMsg), nil
```

No Go build CI ran on this PR, so these type errors were not caught automatically.

### Optional

- **O1:** Non-text artifact parts (URLs, data) are silently dropped. The `message` handler covers URL parts — consider adding parity here, or documenting the exclusion.
- **O2:** `WriteTimeout` 300s is a global cap that will also affect long-lived SSE streams. `http.ResponseController.SetWriteDeadline` per-handler would be more precise. Fine as a follow-up.

### What looks good

- The overall approach (unmarshal artifact, extract text, attach to status event) mirrors the existing `message` case correctly.
- The `broker.go` subscription short-circuit removal fixes a real latent race — if the first call errors, subsequent calls were silently skipping the RPC.
- README YAML correction matches the actual config schema.

PR #1677, opened 2026-09-16
fix(hub): only record web reply affinity when message channel is explicitly web

zahmedsafeer-hub:
## Summary
Prevents the `webChannelBus` observer spoke from hijacking reply affinity on untagged fan-out messages:

- When an agent sends an untagged outbound message (`channel == ""`), `FanOutEventBus.Publish` fans out to all registered spokes, including the observer spoke `webChannelBus`.
- `webChannelBus.Publish` previously called `RecordChannel(..., "web", ...)` unconditionally, overwriting `webchat_conversation_context` with `"web"`. On Turn 2+, `handleAgentOutboundMessage` read back `"web"` from `GetLastChannel` and routed replies exclusively to `web`, skipping external broker plugins such as `a2a-bridge`.
- Guards `RecordChannel` with `if msg.Channel == "web"`.


ptone:
## Code Review: PR #1677

### Required — Missing regression test

The fix is correct and minimal — guarding `RecordChannel` behind `msg.Channel == "web"` prevents untagged fan-out messages from poisoning reply affinity. However, no test exercises the actual bug scenario.

The existing `TestWebChannelBus_Publish_UpdatesState` already had `Channel: "web"` set before this PR and passes identically with or without the fix. A regression test should publish a message with `Channel: ""` and assert no `webchat_conversation_context` row is created:

```go
func TestWebChannelBus_Publish_UntaggedFanout_SkipsAffinity(t *testing.T) {
    store, db := newTestWebChatStore(t)
    defer db.Close()

    bus := NewWebChannelBus(slog.Default(), store)

    msg := &messages.StructuredMessage{
        Version:   messages.Version,
        Sender:    "agent:coder",
        SenderID:  "agent-uuid-1",
        Recipient: "user:alice",
        Msg:       "Hello",
        Type:      messages.TypeInstruction,
        Channel:   "",  // untagged — the bug scenario
    }

    err := bus.Publish(context.Background(), "scion.project.proj1.user.user1.messages", msg)
    require.NoError(t, err)

    var count int
    err = db.QueryRow(`SELECT COUNT(*) FROM webchat_conversation_context`).Scan(&count)
    require.NoError(t, err)
    require.Equal(t, 0, count, "untagged fan-out must not record web affinity")
}
```

### What looks good

- Guard is placed at exactly the right scope — inside `webChannelBus.Publish` where it has access to `msg.Channel`.
- Consistent with the broker inbound handler which already guards `RecordChannel` with `msg.Channel != ""`.
- The comment update accurately describes the rationale.
- Clean, minimal fix for a real production bug. CI passes.

PR #1810, opened 2026-09-21
fix(store): cast Postgres CTE uuid params after #1787 still 42883s

sal-m2026:
Fixes #1634

Follow-up to #1787. Registering `uuid.UUID` with pgx is not enough when Ent
runs raw SQL through `database/sql`: `google/uuid.UUID`'s `driver.Valuer`
returns a **string**, so `$1` in `SELECT $1 AS id` can still be typed as
text. The recursive CTE in `GetEffectiveGroupsForAgent` then fails with
`SQLSTATE 42883` (`uuid = text`) on the join
`gc.parent_group_id = e.id`. Hub fail-closes principal resolution and
maps that to the generic GCP SA-assign 403 on agent-creates-agent.

This casts the placeholders (`$n::uuid`) on Postgres only. SQLite is
unchanged.

- [x] Tests pass

`TestSQLUUIDPh` covers the helper. The existing Postgres integration test
comments now describe the CTE typing, not only pgx registration.

Verified on a Postgres-backed HA Hub: coordinator `scion start` with a
project-default GCP SA now logs `service-account assignment allowed`
(`callerKind: agent`) and the child agent starts. Same path was 42883 on
Hub `main` (including #1787) before this change.

@ptone — this is the remaining 42883 on the agent-creates-agent path after #1787.


sal-m2026:
Closing — opened before we had sign-off to send this to Preston. The branch remains on the fork; we can reopen once approved.

ptone:
looked up sal-m2026 on [signcla/](http://signcla/)

PR #1847, opened 2026-09-23
feat(hub, a2a-bridge): verify external Google bearer tokens and add Gemini Enterprise A2A v0.3 compatibility

bobbymatthews:
## Overview

Integrations fronting the Scion Hub — specifically the Gemini Enterprise (GE) A2A bridge — forward the end user's own Google OAuth/OIDC credential as a standard `Authorization: Bearer` header rather than minting or holding a pre-configured Hub credential.

This PR enables the Hub's `UnifiedAuthMiddleware` to authenticate and provision these external bearer tokens directly against configured trusted issuers and sign-in policies. In tandem, it updates the A2A bridge to act as a pass-through for end-user bearer tokens, removes the interim bridge-side OAuth provider shim, and adds full A2A v0.3 wire protocol compatibility for Gemini Enterprise.

---

### 1. Hub Auth Middleware (`pkg/hub`)
* **External Bearer Token Verification (`auth_external_bearer.go`)**:
  * **Google OAuth2 Access Tokens (`ya29.*`)**: Introspected via Google's `tokeninfo` endpoint and bound to the `accounts.google.com` (`issuer_type: user`) trusted issuer. Enforces that `aud`/`azp` matches the configured `expected_audience` (OAuth Client ID) and that the email is verified. Token validation results are cached in-memory by SHA-256 token hash for `min(expires_in, 5m)`.
  * **OIDC ID Tokens**: Verified via the existing `FederationAuthenticator`.
* **Centralized Policy & Provisioning Enforcement**:
  * Verified identities are provisioned through `ExternalUserProvisioner` using the exact same sign-in governance as interactive OAuth/proxy logins (`admin_emails`, `authorized_domains`, `user_access_mode`).
  * Denied domains and suspended users fail closed with `403 Forbidden` (`user_suspended` / `access denied`); invalid tokens yield `401 Unauthorized`.
  * Requests authenticated via this path carry the `external-bearer` auth-type context.
  * When no trusted Google issuer is configured, the path is inert and default rejection semantics are preserved.

### 2. A2A Bridge Pass-Through & GE Compatibility (`extras/scion-a2a-bridge`)
* **Transparent Token Forwarding**:
  * Updated `UATValidator` to classify external non-PAT bearer tokens as `"bearer"`, allowing `callerHubClient` to forward them directly to the Hub without local exchange.
  * Removed the interim bridge-side OAuth shim (`oauth_provider.go` and `/oauth/*` endpoints) that previously minted super-admin tokens directly against SQLite.
* **Gemini Enterprise A2A v0.3 Compatibility**:
  * Handles v0.3 JSON-RPC / REST method shapes (e.g., `message/send`, agent-addressed `POST /projects/{p}/agents/{a}`).
  * Translates responses into the v0.3 task schema expected by GE clients.
  * Dynamically shapes the agent card for GE callers (`protocolVersion: 0.3.0`, Google OAuth / Bearer schemes, capability negotiation), while serving standard SDK cards to non-GE clients (governed by `Vary: User-Agent` and `Cache-Control: private`).

### 3. Documentation (`docs-site`)
* Added the **External Bearer Tokens** guide to `docs-site/src/content/docs/hosted/single-node/auth.md`, documenting configuration parameters (`trusted_issuers`, `expected_audience`), token flow, and policy resolution.

---

## Security & Governance Considerations

* **Audience Binding**: Requiring `expected_audience` on opaque Google access tokens prevents token-confusion attacks where any valid Google access token (e.g., from `gcloud`) could otherwise authenticate its holder.
* **No Bridge Privilege Escalation**: Eliminates the risk of the bridge minting administrative Hub tokens or bypassing Hub-level role gating.
* **Fail-Closed Policy**: If external tokens cannot be verified against a configured trusted issuer, or if the user is suspended or domain-restricted, requests are immediately denied at the Hub middleware layer.

---

## Reconciliation Note (PR #1741)

This branch has been rebased on current `main` (`30ca5b927`), cleanly reconciling with upstream PR #1741 (*"Google credential exchange and Gemini Enterprise A2A compatibility"*):
* Resolved naming collisions between token validators (`emailOnlyUserStore` test fake, shared `googleTokenInfoURL`).
* Preserved upstream REST handler agent-card emission while scoping GE-specific card customizations via `isGEClient`.

---

## Verification & Testing

* **Build**:
  ```bash
  go build -buildvcs=false ./...


ptone:
## Code Review: PR #1847

### Critical — Privilege escalation via JWT external-bearer branch (C1)

**`pkg/hub/auth_external_bearer.go`** — The JWT branch in `authenticateExternalBearer` accepts tokens from **any** configured trusted issuer (including `service_account` issuers and non-Google user issuers like Firebase) and provisions them as full Hub users via `provisionUser`. This grants **admin** if the email matches `admin_emails`, and never checks `email_verified`.

The same token presented via `X-Scion-Federation-Token` (existing federation path) would only get a scope-less viewer with admin→viewer role cap. This is a privilege escalation.

**Attack scenario:** A Hub trusts a Firebase issuer that allows unverified email sign-up. An attacker signs up as `admin@corp.com`, sends the resulting ID token as `Authorization: Bearer`, and is provisioned as Hub admin.

Additionally, there is no `sub` binding (unlike #1741's `resolveLocalUser`), so an email reassigned at the IdP inherits the Hub account.

**Fix:** Restrict the JWT branch to the Google user issuer only (`issuer_type: user`), require `email_verified == true`, reject `FederatedServiceIdentity`. Better: reuse `GoogleCredentialValidator` + #1741's resolve/bind logic.

### Required — Duplicates merged #1741 (R1)

`main` already has `GoogleCredentialValidator`, `GEExchangeService` (sub-bound, member-only provisioning, rate-limited), the `geGoogle` bridge scheme, and a v0.3 REST handler. This PR adds a parallel set: second tokeninfo verifier, second provisioning policy (email-only, admin-capable), second bridge auth route, second v0.3 translation — with weaker invariants.

**Fix:** Build on `GoogleCredentialValidator` and `GEExchangeService.resolveLocalUser` so both paths share one verifier and one binding/provisioning policy. Explain why the exchange doesn't meet the GE need, or use it.

### Required — Bridge logs raw request/response bodies (R2)

`s.log.Info("incoming A2A JSON-RPC request", ..., "raw_body", string(bodyBytes))` writes full user prompts and agent replies to production logs for every JSON-RPC request. This is PII and potentially contains secrets.

**Fix:** Remove both body logs, or log only method/id/size at Debug level.

### Required — v0.3 detection misclassifies v1.0 requests (R3)

`bytes.Contains(bodyBytes, "\"kind\"")` matches anywhere in the body, including user message text. When it matches, a v1.0 client's response is rewritten into v0.3 shape. gemini-code-assist raised the same point (unaddressed).

**Fix:** Decide `isV03` from the parsed `method` field only, not byte-sniffing.

### Required — v0.3 task-ID handling breaks multi-turn (R4)

`taskId` is stripped so a2a-go creates a new task each time. The response ID is overwritten with the client's ID, which the server never stored. Follow-up `tasks/get`/`tasks/cancel` return not-found. Streaming methods are rewritten to non-streaming `SendMessage` and returned as buffered JSON.

**Fix:** Keep a clientTaskID↔serverTaskID map, or reject unknown task continuations. Return JSON-RPC error for streaming methods or implement them.

### Required — No tests for ~300 new bridge lines (R5)

No test coverage for `convertResponseToA2AV03`, request rewriting, `isGEClient` card shaping, `bufferedResponseWriter`, `hubUAT` non-PAT token handling, or the JWT branch where C1 lives.

### Required — Agent card advertises Google OAuth for every non-`none` scheme (R6)

The `securitySchemes` block is added whenever `cfg.Auth.Scheme != "none"`, so `apiKey`, `bearer`, `federation` etc. all publish an auth method they don't accept. `bearerFormat: JWT` is wrong even for GE (which sends opaque `ya29` tokens).

**Fix:** Emit the scheme only when configured for Google pass-through.

### Optional / Nit

- **Unauthenticated amplification:** Every unknown `ya29.` token triggers an outbound call to Google tokeninfo (up to 5s). No negative caching, no rate limiting (unlike #1741's exchange limiter). Add a per-IP limiter.
- **Nit:** Send access token to tokeninfo as POST form body, not `?access_token=` in URL (keeps it out of proxy/access logs).
- **Nit:** `isGEClient` = `strings.Contains(UA, "Google")` also matches Googlebot, Google-Apps-Script, etc.
- **Nit:** `_ = contextID` is dead code.
- **Nit:** `tokenInfoCache` is a package global shared across tests and instances.

### What looks good

- The `ya29` path is well-built: mandatory audience binding, `email_verified` checked, `allowed_emails` enforced, policy re-checked on cache hits.
- Refusing opaque tokens when `expected_audience` is unset closes the token-confusion hole.
- 403 (policy/suspended) vs 401 (bad token) split preserves existing error semantics.
- `Cache-Control: private` + `Vary: User-Agent` on the client-dependent card is correct.
- The 11 hub tests for the `ya29` path are thorough.

ptone:
closing in favor of new plan in https://github.com/ptone/scion/issues/1816

